PR #17, opened 2015-03-27
fix utf8 inconsistencies

kergma:
There are some inconsistencies in pg_enable_utf8 behaviour, described in cpan bugs page 
https://rt.cpan.org/Public/Bug/Display.html?id=103137

this will fix them

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -1184,10 +1184,11 @@ SV * dbd_st_FETCH_attrib (SV * sth, imp_sth_t * imp_sth, SV * keysv)
 			SV * sv_fieldname;
 			retsv = newRV_inc(sv_2mortal((SV*)av));
 			while(--fields >= 0) {
+				D_imp_dbh_from_sth;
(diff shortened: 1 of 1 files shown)


nwellnhof:
Why isn't this fixed in over a year? This is a serious issue for legacy setups with `pg_enable_utf8=>0`.

machack666:
I am reviewing this patch now and will merge if everything passes.

machack666:
This has caused failing tests with some of the test data in t/30unicode.t.  I will investigate further and see if I can find out why and if it's to do with this patch or if it was already failing.

machack666:
Yeah, it's definitely the patch causing the failures.  I'll see what I can come up with, or if the original author want to take a look at this with the latest implementation that'd be great.

machack666:
@kergma ^^^

machack666:
After some analysis, we've concluded the test that is failing is bad.  I'm working on getting a better test given the patch here and we'll get this committed.

PR #23, opened 2017-04-04
Support binary COPY format

ilmari:
- Use the length returned by PQgetCopyData when storing it in the SV
- Clear the UTF-8 flag, since it's binary data.

Also make sure UTF8 flag is correct in pg_getcopydata in non-binary mode. We're replacing the contents of an existing SV, so make sure the UTF8 flag is set correctly regardless of what it was before.


ilmari:
I just noticed `pg_putcopydata` is also broken, so probably don't merge this until I've fixed that too.

PR #35, opened 2018-02-26
Eliminate this warning:

jkeenan:
Useless use of concatenation (.) or string in void context at Makefile.PL line
275.

From the context, it appears string concatenation was intended.


ilmari:
This is a duplicate of https://rt.cpan.org/Public/Bug/Display.html?id=127097, which was fixed in commit 8010dcd771e5902e867c24c18589d095b3f6b421.

PR #43, opened 2018-12-31
Drop support for PostgreSQL before 8.0

ilmari:
We currently (since 2007) support PostgreSQL back to version 7.4, which has been out of
support upstream since 2010.  By dropping support for versions before 8.0 we can get rid
of several bits of cruft:

- PQprepare() emulation
- PQserverVersion() emulation
- "smart" pg_server_prepare mode
- conditional catalog SQL query fragments

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,9 @@
 
 Version 3.8.0
 
+  - Increase minimum supported PostgreSQL version to 8.0
+    [Dagfinn Ilmari Mannsåker]
+
   - Add support for foreign tables in table_info() and column_info()
     [Dagfinn Ilmari Mannsåker]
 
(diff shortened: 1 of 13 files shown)


turnstep:
My opinion is the same: support as far back as possible, until it becomes
an undue burden or technically infeasible to support specific versions. But
I'm fine with dropping 7.4 support. :)

ilmari:
I take that as an okay to merge this PR?

On a related note we really should set up some CI (Travis, Circle, whatever) to automatically test on every version we claim to support, so we don't accidentally break them again.

turnstep:
Yes, I'm fine with the merge. Some sort of CI would certainly be nice,
although I'm out of tuits.

PR #48, opened 2019-06-07
Proof of concept for ->pg_error_field()

turnstep:


pali:
Hi! Have you seen discussion about Better diagnostic support in DBI?
https://www.nntp.perl.org/group/perl.dbi.dev/2019/01/msg8204.html

turnstep:
I did see that thread, and I think it's an interesting idea although I would think both pg_error_field() and get_diagnostics() could co-exist.

As far as this from Tim:
```
I'd much prefer to wait till multiple drivers have added their own
driver-specific, and driver-optimized, interface.
```
I really don't see that happening. I say we drive forward, rather than waiting for other drivers and getting official DB blessings.

turnstep:
Closing: added n version 3.9.0

PR #50, opened 2019-06-27
Fix encoding of SQL_VARBINARY type in $dbh->quote() function

pali:
For SQL_VARBINARY DBI type use same logic as for PG_BYTEA type.
SQL_VARBINARY DBI type is just alias for PG_BYTEA type, so result of
quoting operation should be same.

Prior this patch following code

    $dbh->quote("\xC3\xA1", SQL_VARBINARY)

returned

    E'\\303\\203\\302\\241'

And

    $dbh->quote("\xC3\xA1", { pg_type => PG_BYTEA })

returned

    E'\\303\\241'

Now with this patch both PG_BYTEA and SQL_VARBINARY return same value

    E'\\303\\241'

which is correct.

### DIFF
--- a/Pg.xs
+++ b/Pg.xs
@@ -283,7 +283,7 @@ quote(dbh, to_quote_sv, type_sv=Nullsv)
 			}
 
 			/* At this point, type_info points to a valid struct, one way or another */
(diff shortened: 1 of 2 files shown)


machack666:
Looks good, thanks.

Issue #44, opened 2019-01-14
ShowErrorStatement doesn't work on do("") requests


dk:
Please compare the outputs of the two requests that should be identical, both with 

`$dbh->{ShowErrorStatement} = 1;`

pre-executed, both sending a text to an integer argument:

`my $sth = $dbh->prepare('select relname from pg_class where relpages = ?');
$sth->execute('foo');`

`DBD::Pg::st execute failed: ERROR:  invalid input syntax for integer: "foo" [for Statement "select relname from pg_class where relpages = ?" with ParamValues: 1='foo'] `

and

`$dbh->do('select relname from pg_class where id relpages ?', {}, 'bar');`

`DBD::Pg::db do failed: ERROR:  syntax error at or near "relpages"
LINE 1: select relname from pg_class where id relpages $1
                                              ^ [for Statement "select relname from pg_class where id relpages ?"]`

The latter is missing the ParamValues explained.

turnstep:
At first glance, I don't know if this is possible. The problem is that DBI wants to call ParamValues on the current handle. For statement handles, this is no problem, as the placeholders are stored in the sth structure (which ParamValues walks through via dbd_st_FETCH_attrib). For database handles, however, we create a very local and temporary imp_sth inside of Pg.xs, which goes out of scope well before DBI asks $dbh for ParamValues (which it cannot handle anyway, as it knows nothing about it). We could in theory jump through some hoops to have imp_dbh start storing detailed information from its temporary imp_sth creation, but seems a lot of trouble for something so trivial.

I will expand the documentation to make this clearer, however.

dk:
It would help debugging a lot to have parameters dumped together with the error. Or at the very least fix the documentation with example how to rewrite do() calls into something that respects ShowErrorStatement..

dk:
I've sent a pull request please review. Please disregard d636fc6 , it's still there because only github didn't roll it back when I did 'git reset'.

turnstep:
Added as of version 3.9.0

PR #52, opened 2019-07-02
refs #44: show ParamValues also when do() syntax is used

dk:
Hi,

Here I'm trying to show ParamValues in do() calls, that wasn't possible before as do() and fetch(ParamValues) was done outside in the DBI.xs. I'm using the temporary nature of $sth created inside DBD::Pg::db::do(), that is cleaning up link between $sth and $dbh needed to display the ParamValues. 

Hope that makes sense
/dk


dbdimp.c
```diff
@@ -685,6 +685,15 @@ void dbd_db_destroy (SV * dbh, imp_dbh_t * imp_dbh)
...
 
 	if (TSTART_slow) TRC(DBILOGFP, "%sBegin dbd_db_destroy\n", THEADER_slow);
+ 
+	if ( imp_dbh->do_tmp_sth != NULL) {
+		imp_sth_t * imp_sth = imp_dbh->do_tmp_sth;
+		while ( imp_sth != NULL && imp_sth->do_tmp_owner == imp_dbh ) {
+			imp_sth->do_tmp_owner = NULL;
+			imp_sth = imp_sth->do_tmp_old_val;
+		}
+		imp_dbh->do_tmp_sth == NULL;
```
turnstep: s/==/=/

Pg.xs
```diff
@@ -356,6 +356,9 @@ void do(dbh, statement_sv, attr=Nullsv, ...)
 			if (!dbdxst_bind_params(sth, imp_sth, items-2, ax+2))
 				XSRETURN_UNDEF;
 			imp_sth->async_flag = asyncflag;
+			imp_sth->do_tmp_owner = imp_dbh;
+			imp_sth->do_tmp_old_val = imp_dbh->do_tmp_sth;
```
turnstep: Do we really need to store the previous sth? Each dbh should create exactly one temporary sth at a time.

dbdimp.c
```diff
@@ -778,12 +787,14 @@ SV * dbd_db_FETCH_attrib (SV * dbh, imp_dbh_t * imp_dbh, SV * keysv)
...
 
-	case 11: /* pg_INV_READ  pg_protocol */
+	case 11: /* pg_INV_READ  pg_protocol ParamValues */
 
 		if (strEQ("pg_INV_READ", key))
 			retsv = newSViv((IV)INV_READ);
 		else if (strEQ("pg_protocol", key))
 			retsv = newSViv((IV)imp_dbh->pg_protocol);
+		else if (strEQ("ParamValues", key) && imp_dbh->do_tmp_sth != NULL)
+			return dbd_st_FETCH_attrib (dbh, imp_dbh->do_tmp_sth, keysv);
```
turnstep: Heh, nice trick

dbdimp.c
```diff
@@ -3935,6 +3947,11 @@ void dbd_st_destroy (SV * sth, imp_sth_t * imp_sth)
 	if (NULL == imp_sth->seg) /* Already been destroyed! */
 		croak("dbd_st_destroy called twice!");
 
+	if ( imp_sth->do_tmp_owner != NULL && imp_sth->do_tmp_owner->do_tmp_sth == imp_sth) {
```
turnstep: We can probably take this out completely if we don't worry about "other" sths.

dbdimp.c
dk: yes

Pg.xs
dk: Most probably no, but I'm being paranoid about areas I don't know. Namely, could be an execute() call inside execute()? If no, then let's remove it

dk:
Other than this looks fine. How do we proceed from here? Could you apply the revised code, or should I file a new or updated pull request?

turnstep:
Let's keep going the git(hub) way: make some changes to this branch and then push it out. Three things I think need addressing:
* Simplify the code and don't worry about saving/restoring previous sth. It's very short-lived and there will never be more than one.
* Pg.pm docs
* That one typo :)

dk:
Kindly check out if it's okay now

turnstep:
Coming along nicely! Still needs a ParamValues inside the Database Handle Attribute section.

turnstep:
Also give yourself credit in the Changes file!

dk:
Here's some more, please take a look

greg-binary:
If you can clear up the conflict with the "Changes" file, I can merge this in.

dk:
I did merge the changes for 3.8.1. Github still though shows "conflict".

PR #55, opened 2019-08-17
Support partitioned tables with pg_class.relkind = 'p'

octaviancorlade:
When partitioned tables exist `$dbh->table_info()` is currently listing all the partitions (which are stored as tables, so it makes sense) but it does not include the parent partitioned tables, introduced in Postgres 10 with `relkind = 'p'`.

An alternative approach could be to create a new `TABLE_TYPE` value for partitioned tables.


turnstep:
Thanks for writing this. I think it's fine as is, no need for a new TABLE_TYPE for these.

PR #58, opened 2019-08-20
Add DBI SQL_BLOB, SQL_BINARY and SQL_LONGVARBINARY types as alias for PG_BYTEA

pali:
DBI type SQL_VARBINARY is already aliased to PG_BYTEA. Other DBI drivers
use these DBI types for binding binary data so also add them to DBD::Pg.


machack666:
Looks good, thanks.

Issue #37, opened 2018-04-20
Hard to trigger crash with async queries and InactiveDestroy


johto:
Hi,

I unfortunately can't share the test case since it uses thousands of lines of proprietary code, but it does roughly this:

```Perl
use DBI;
use DBD::Pg ':async';
use Time::HiRes;

sub mayhem {
    my $dbh = DBI->connect("dbi:Pg:", '', '', {pg_enable_utf8 => 1, PrintError => 0, RaiseError => 1, pg_server_prepare => 0});
    my $sth = $dbh->prepare(q{SELECT CASE WHEN pg_sleep(random() * 0.25) IS NOT NULL THEN random()::text::uuid END}, {pg_async => PG_ASYNC});
    $dbh->{InactiveDestroy} = 1;
    $sth->execute();
    for (my $i = 0; ; ++$i) {
        if ($sth->pg_ready) {
            last;
        }
        if ($i >= 100) {
            print STDERR "wait timeout\n";
            $sth->cancel();
            exit(1);
        }
        Time::HiRes::sleep(0.05);
    }
    eval {
        $sth->pg_result();
    };
}
mayhem();
```

which sometimes crashes with a backtrace like this:

```
(gdb) bt
#0  0x00007ffff61c40fb in PQclear () from /usr/lib/x86_64-linux-gnu/libpq.so.5
#1  0x00007ffff64107f8 in pg_db_destroy (dbh=0x1d1bc40, imp_dbh=0x1d22570) at dbdimp.c:688
#2  0x00007ffff6407b54 in XS_DBD__Pg__db_DESTROY (my_perl=0x603010, cv=<optimized out>) at ./Pg.xsi:390
#3  0x00007ffff664a656 in XS_DBI_dispatch (my_perl=<optimized out>, cv=0x8c0268) at DBI.xs:3797
#4  0x00007ffff7b0ad26 in Perl_pp_entersub () from /usr/lib/libperl.so.5.18
#5  0x00007ffff7a94846 in Perl_call_sv () from /usr/lib/libperl.so.5.18
#6  0x00007ffff7b13d71 in ?? () from /usr/lib/libperl.so.5.18
#7  0x00007ffff7b14500 in Perl_sv_clear () from /usr/lib/libperl.so.5.18
#8  0x00007ffff7b14b8d in Perl_sv_free2 () from /usr/lib/libperl.so.5.18
#9  0x00007ffff7aecf48 in ?? () from /usr/lib/libperl.so.5.18
[...]

PR #60, opened 2019-09-27
Fix a segfault related to async statements

chazmcgarvey:
Hi friends,

After upgrading to DBD::Pg 3.10.0, my async app now dies on "Segmentation fault" or "double free or corruption". The exact place it dies is inconsistent, but capturing various backtraces reveals it's always in `PQclear`. So it seems clear we're attempting to free a result more than once. Here's a sample backtrace:

    #0  0x00007fe110a45a90 in free () from /lib64/libc.so.6                                                               
    #1  0x00007fe10d30f9fe in PQclear () from /usr/lib64/postgresql-11/lib64/libpq.so.5                                   
    #2  0x00007fe10d358089 in _result () from .../lib/perl5/x86_64-linux/auto/DBD/Pg/Pg.so
    #3  0x00007fe10d35994c in pg_st_deallocate_statement.isra () from .../lib/perl5/x86_64-linux/auto/DBD/Pg/Pg.so
    #4  0x00007fe10d365023 in pg_st_destroy () from .../lib/perl5/x86_64-linux/auto/DBD/Pg/Pg.so
    #5  0x00007fe10d352ea6 in XS_DBD__Pg__st_DESTROY () from .../lib/perl5/x86_64-linux/auto/DBD/Pg/Pg.so
    #6  0x00007fe10d3939b8 in XS_DBI_dispatch () from .../lib/perl5/x86_64-linux/auto/DBI/DBI.so
    #7  0x000055c2a3321179 in Perl_pp_entersub ()
    ...

I bisected to b312efbfd7e605286f00ad373d42c9b2e6b0c13d (made prior to 3.9.0).

I took a stab at fixing the problem and came to this solution. This *does* alleviate the segfaults for me, though I'd need someone more familiar with this code to verify it's the right solution.

Although it crashes consistently in my complex app, I haven't yet succeeded in even reproducing the crash in a test file. :-(

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -684,6 +684,8 @@ void dbd_db_destroy (SV * dbh, imp_dbh_t * imp_dbh)
 		if (imp_dbh->async_sth->result) {
 			TRACE_PQCLEAR;
 			PQclear(imp_dbh->async_sth->result);
+			if (imp_dbh->last_result == imp_dbh->async_sth->result)
+				imp_dbh->last_result = NULL;
 		}
 		imp_dbh->async_sth = NULL;
 	}
@@ -5169,8 +5171,13 @@ long pg_db_result (SV *h, imp_dbh_t *imp_dbh)
 
 		if (imp_dbh->async_sth) {
 			if (imp_dbh->async_sth->result) { /* For potential multi-result sets */
-				TRACE_PQCLEAR;
-				PQclear(imp_dbh->async_sth->result);
+				if (imp_dbh->sth_result_owner == (long int)imp_dbh->async_sth) {
+					imp_dbh->sth_result_owner = 0;
+				}
+				else {
+					TRACE_PQCLEAR;
+					PQclear(imp_dbh->async_sth->result);
+				}
 			}
 			imp_dbh->async_sth->result = result;
 		}


dbdimp.c
```diff
@@ -684,6 +684,8 @@ void dbd_db_destroy (SV * dbh, imp_dbh_t * imp_dbh)
 		if (imp_dbh->async_sth->result) {
 			TRACE_PQCLEAR;
 			PQclear(imp_dbh->async_sth->result);
+			if (imp_dbh->last_result == imp_dbh->async_sth->result)
```
chazmcgarvey: Actually this change does nothing to alleviate segfaults for me, but looks safer to me to do this...

a6502:
Hi,

I just ran into the same double free issue in an application that uses async extensively and can confirm that the above patch fixes that.

I don't have a small test-case but at least I can show the code: https://github.com/a6502/jobcenter/blob/master/bin/maestro

This uses Mojo::Pg (subclassed for better connection pooling).

Is there anything I can do to help solve this issue?

chazmcgarvey:
Yeah, my app also uses Mojo::Pg.

I considered that maybe Mojo::Pg was misusing DBD::Pg somehow, but after digging through the the dbdpg I settled on the flaw being here. What do your think @turnstep ?

johto:
Maybe a duplicate of #37?

chazmcgarvey:
Looks very similar, but I didn't think it was the exact same bug because I bisected this crash to a commit made later than that bug was filed (b312efb).

esabol:
Merge conflict. Needs rebasing?

chazmcgarvey:
> Merge conflict. Needs rebasing?

Actually this PR should be abandoned because the issue was fixed with the 3.10.1 release.

Issue #63, opened 2019-11-08
Time::Piece handling


mla:
This script runs fine with DBD::Pg 3.5.3, whether you uncomment the use Time::Piece line or not.

```perl
#!/usr/bin/env perl

use strict;
use Carp;
use DBI;
use DBD::Pg;
#use Time::Piece;

my $dbh = DBI->connect('dbi:Pg:dbname=try', undef, undef)
  or die 'db connection failed';

$dbh->do(q{
  DROP TABLE IF EXISTS mytable
});

$dbh->do(q{
  CREATE TABLE mytable (
    creation_date date
  )
});

my $now = localtime;
$dbh->do(q{
  INSERT INTO mytable (creation_date) VALUES (?)
}, undef, $now);
```

With 3.6.0, it fails if Time::Piece is loaded:

```
$ ./try
DBD::Pg::db do failed: ERROR:  invalid input syntax for type date: "{"48","8","21","7","10","119","4","310","0","1573189728","1"}" at ./try line 23.
```

```
$ perl -v

This is perl 5, version 18, subversion 2 (v5.18.2) built for i686-linux
```

I'm guessing this is perhaps related to this change?
```
 - Make sure we do not inadvertently modify the string passed to prepare() when
 
    doing the new backslash escape manipulation.
    [Greg Sabino Mullane]
    (CPAN ticket #114000)
```

I'm not sure what the preferred behavior is? We can require manual stringification of the Time::Piece object, of course. Just was rather nice that the old behavior stringified it automatically and required no change to the code.

mla:
I see the same behavior with the latest DBI and DBD::Pg, of course:

```
# DBI                         Version 1.642
# DBD::Pg                     Version 3.10.0
# Perl                        Version 5.18.2
# OS                          linux
# PostgreSQL (compiled)       90614
# PostgreSQL (target)         90614
# PostgreSQL (reported)       PostgreSQL 9.6.14 on i686-pc-linux-gnu, compiled by gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23), 32-bit
```

mla:
Strangely though, on another system, the same script does *not* fail:

```
# DBI                         Version 1.642
# DBD::Pg                     Version 3.10.0
# Perl                        Version 5.18.2
# OS                          linux
# PostgreSQL (compiled)       90615
# PostgreSQL (target)         90615
[...]

PR #64, opened 2019-11-09
fix test on 32-bit systems and problem with blessed placeholder values

mla:
A test was failing on 32-bit systems related to PG_MIN_BIGINT:

>t/04misc.t .. 1/101 DBD::Pg::st execute failed: ERROR:  invalid input syntax for integer: "-9.22337203685478e+18" at t/04misc.t line 87.

I'm not sure what the best solution is. I just added a "use bignum" to that test file. Looks like bignum was added to core in 5.8. I don't know if that's acceptable or not.

Also, there was a problem with handling blessed objects (see my example on issue #63).
It appeared to be due to the ordering of tests, where a Time::Piece instance was being treated as an array ref. I reordered that section a bit.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2482,23 +2482,25 @@ int dbd_bind_ph (SV * sth, imp_sth_t * imp_sth, SV * ph_name, SV * newvalue, IV 
 			currph->iscurrent = DBDPG_TRUE;
 			imp_sth->has_current = DBDPG_TRUE;
 		}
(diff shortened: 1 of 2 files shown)


turnstep:
I think these issues were fixed in version 3.10.2 (32-bit tests) and 3.10.4 (Time::Piece handling). Could you confirm this is the case so we could possibly close this pull? Thanks!

turnstep:
Closing, as I am fairly confident this issue has been fixed. Please reopen if that is not the case.

PR #75, opened 2020-10-27
Replace use of "vars" with "our"

JRaspass:
Use of this pragma is discouraged in favour of "our" under v5.6+ and this dist depends on v5.8+.

my rather than our in Makefile.PL seemed more logical.


turnstep:
Looks good, but could you rebase? It's complaining about some whitespace in README.dev.  There are also a few minor complaints from t/99_perlcritic.t when I tried it out.

JRaspass:
@turnstep rebased and fixed the Critic violations, my bad I wasn't using `AUTHOR_TESTING=1` before.

PR #76, opened 2020-11-05
Switch to XSLoader

toddr:
We patch DBD::Pg to use the lighter XSLoader in our perl. I'm including the patch in case you want to use it.

### DIFF
--- a/Pg.pm
+++ b/Pg.pm
@@ -19,10 +19,10 @@ use 5.008001;
     use version; our $VERSION = qv('3.15.0');
 
     use DBI ();
-    use DynaLoader ();
+    use XSLoader ();
     use Exporter ();
     use vars qw(@ISA %EXPORT_TAGS $err $errstr $sqlstate $drh $dbh $DBDPG_DEFAULT @EXPORT);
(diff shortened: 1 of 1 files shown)


turnstep:
Thanks! Applied a version of this in d8f4bd1bb0eca1821df34470735e25c2b412577c

ilmari:
Given that `XSLoader` is a core module since perl 5.6, do we really need the fallback?

turnstep:
*shrug* I suppose not, but the fallback code is only a few lines, so I am on the fence.

toddr:
> _shrug_ I suppose not, but the fallback code is only a few lines, so I am on the fence.

The comment I would make about your commit is that if either of those modules fail to load, they now fail silently.

turnstep:
I don't think it would be silent, it would give: `Can't locate object method "bootstrap" via package "DBD::Pg" at blib/lib/DBD/Pg.pm line 105`

turnstep:
Anyway, I am fine now with yanking out Dynaloader: we can always add it back in easy enough.

PR #85, opened 2021-10-20
Automatically use 64-bit versions of large object functions when avallable

ilmari:
This mostly reverts commit 751f5ca5ef664f0af5934db34d7e217cd4a61376 and replaces it with an approach originally submitted by @pgguru in 2017. The 64-suffixed method names are retained as backwards compatibility aliases.

Also add tests for LOs of size INT_MAX and larger. Large objects are stored sparsely, so this doesn't actually take any significant space in the database.

In passing, convert the test script to use `done_testing()`.

### DIFF
--- a/Changes
+++ b/Changes
@@ -22,6 +22,9 @@ Version ?? (unreleased)
  - Use non-root user when calling pg_resetwal
    [Greg Sabino Mullane]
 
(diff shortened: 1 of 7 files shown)


esabol:
@ilmari : Looks like there are merge conflicts on this PR. Can you resolve? Thanks!

PR #88, opened 2021-12-08
Fix reading binary data from ParamValues

dur-randir:
In our DBI plugin that make DB reconnect seamless for an end user, we transfer current to-be executed query parameters from an old connection to the new one[0]. But this fails in case when some data in parameters was binary. This PR fixes this case by not relying on strlen() calculations but using already known string length instead.

[0] https://metacpan.org/dist/DBIx-RetryOverDisconnects/source/lib/DBIx/RetryOverDisconnects.pm#L550

### DIFF
--- a/t/02attribs.t
+++ b/t/02attribs.t
@@ -18,7 +18,7 @@ my (undef,undef,$dbh) = connect_database();
 if (! $dbh) {
     plan skip_all => 'Connection to database failed, cannot continue testing';
 }
-plan tests => 284;
+plan tests => 285;
 
 isnt ($dbh, undef, 'Connect to database for handle attributes testing');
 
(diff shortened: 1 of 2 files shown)


t/02attribs.t
```diff
@@ -856,6 +856,14 @@ $sth->execute();
 $attrib = $sth->{ParamValues};
 is_deeply ($attrib, $expected, $t);
 
+$t='Statement handle attribute "ParamValues" works with NULL-embeded strings';
```
esabol, line 859: You misspelled "embedded" here. Please fix.

esabol (changes requested):

t/02attribs.t
dur-randir: done

esabol (approved):

dur-randir:
@turnstep Can you please take a look at this? It can result in a data loss, and the fix is just a one line change. I'd greatly appreciate if it gets released.

Issue #98, opened 2022-09-05
Hanging execution of queries like "COPY TO STDIN" in asynchronous mode.


true-alex:
When calling "COPY ... TO STDIN" query in asynchronous mode after pg_ready returned 1, pg_result call hangs.

The reason is probably the implementation of the pg_db_result method in the dbdimp.c file. This method is written exactly with the libpq documentation, the async section, which says that PQgetResult must be called until it returns null. However, if PQgetResult returned a PGRES_COPY_* state, this state will not change until the copy operations are completed (according to the copy libpq documentation).

To test my theory, at the end of the main loop of the pg_db_result function, I put a check with a loop break:
```
diff -r DBD-Pg-3.15.1/dbdimp.c DBD-Pg-3.15.1_patch/dbdimp.c
5375a5376,5378
>         if ( rows == -1 ) {
>               break;
>         }
```
In the attachments there are two greatly simplified files, one synchronous one works correctly, the second asynchronous one without a patch freezes, with a patch it works correctly.

The problem has been reproduced on:
Freebsd 9.4 + Postgress 9.3
Cenos 6 + Postgress 9.4
OracleLinux 8 + Postgress 13
Freebsd 13 + Postgress 13

[test.tgz](https://github.com/bucardo/dbdpg/files/9491191/test.tgz)

true-alex:
When researching the problem, I used the documentation:
https://postgrespro.com/docs/postgrespro/13/libpq-copy
_The functions of this section should be executed only after obtaining a result status of PGRES_COPY_OUT or PGRES_COPY_IN from [PQexec](https://postgrespro.com/docs/postgrespro/13/libpq-exec#LIBPQ-PQEXEC) or [PQgetResult](https://postgrespro.com/docs/postgrespro/13/libpq-async#LIBPQ-PQGETRESULT)._
...
_After [PQgetCopyData](https://postgrespro.com/docs/postgrespro/13/libpq-copy#LIBPQ-PQGETCOPYDATA) returns -1, call [PQgetResult](https://postgrespro.com/docs/postgrespro/13/libpq-async#LIBPQ-PQGETRESULT) to obtain the final result status of the COPY command. One can wait for this result to be available in the usual way. Then return to normal operation._

https://postgrespro.com/docs/postgresql/13/libpq-async
[...]

PR #101, opened 2022-09-29
Fixed hangup of execution of queries like "COPY TO STDIN" in asynchronous mode.

true-alex:
The problem is in the wrong sequence of calls when processing the result of calling the COPY command in asynchronous mode.
https://github.com/bucardo/dbdpg/issues/98

The problem is that the pg_db_result method tries to get the result of the query before it completes. The result of the COPY command will be only after the end of reading / writing the data stream.

Perhaps the problem can be solved differently, I did not want to change the source code much.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -5318,6 +5318,9 @@ long pg_db_result (SV *h, imp_dbh_t *imp_dbh)
             TRACE_PQCLEAR;
             PQclear(result);
         }
+        if ( rows == -1 ) {
+            break;
+        }
     }
 
     if (NULL != imp_dbh->async_sth) {


turnstep:
Do you have a working test script that demonstrates this? While digging in, I found some other issues with regards to async COPY so I want to make sure I fix it all at once.

true-alex:
A lot of time has passed, the server is already down.
According to the results of experiments, I refused to use asynchronous COPY due to the fact that it did not speed up, but only saved memory (I had enough memory).
Only test cases remained, which are also tied to task #98.
https://github.com/bucardo/dbdpg/files/9491191/test.tgz

turnstep:
Thank you. I think we are simply going to disallow calling pg_result after an asynchronous COPY, as it's not the same as the non-COPY cases where we can simply wait for the server to finish.

true-alex:
In the current version of the module, if pg_result is not called after a successful pg_ready, an error will occur, since pg_result sets the status of waiting for data from COPY.

PR #102, opened 2022-09-29
Fixed call pg_pid_number

true-alex:
Problem: The code block returning pg_pid_number could never be called.

Fixed bug added to commit 1bfec5fc2532a2f217ceaf0fa188d8bc977b0154 on June 23, 2007

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -767,14 +767,12 @@ SV * dbd_db_FETCH_attrib (SV * dbh, imp_dbh_t * imp_dbh, SV * keysv)
         }
         break;
 
-    case 10: /* AutoCommit  pg_bool_tf  pg_pid_number  pg_options */
+    case 10: /* AutoCommit  pg_bool_tf  pg_options */
 
         if (strEQ("AutoCommit", key))
             retsv = boolSV(DBIc_has(imp_dbh, DBIcf_AutoCommit));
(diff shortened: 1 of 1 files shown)


true-alex:
```
use strict;
use warnings;
use DBD::Pg;

my $dbh = DBI->connect("dbi:Pg:dbname=postgres;host=localhost;port=5432",
                       "username","",{AutoCommit => 1, RaiseError => 1, PrintError => 0 });
print $dbh->{pg_pid_number}."\n";
$dbh->disconnect();
```

On the pre-patch version, the result is "Use of uninitialized value in concatenation (.) or string at - line 7."

Returns the PID after applying the patch.

I cannot think of a case when, with careful programming, this call would be necessary, except for debugging, it was not for nothing that this call was in the code with the comment "undocumented". Unless to understand after the fork the connection is open in this process or in parent.

esabol:
@true-alex : I mean add a test to one of the `t/*.t` files as part of your pull request. Thank you!

turnstep:
On digging into this, it is simply getting the current PID, which is already available to programs via $$ or getpid(), so maybe we don't need to expose this at all? It's only in the code as a convenience for building the name of a prepared statement.

turnstep:
Removed mention of it in bd8a349b33aa4080232f6a0380d8f0d5ebcb7dad

PR #112, opened 2023-03-14
Allow MYMETA.* files to be generated

oalders:
These files are needed for installation via App::cpm. See
https://github.com/skaji/cpm/issues/234 This reverts a change which was
introduced in 7856005dd7437ef7b0

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -232,7 +232,6 @@ my %opts =
      PERL_MALLOC_OK => 1,
      NEEDS_LINKING  => 1,
      NO_META        => 1,
-     NO_MYMETA      => 0,
      NORECURS       => 1,
      PM             => {
                         'Pg.pm' => '$(INST_LIBDIR)/Pg.pm',


perlpunk:
Yes, we also need these files on openSUSE to generate the spec file with requirements.

oalders:
@theory thoughts on this PR?

theory:
Uhhh…no? Seems reasonable. Can't think of why we'd want to prevent there creation of `MYMETA` files.

turnstep:
Closing as done elsewhere - sorry this took so long to get on my radar!

perlpunk:
@turnstep I just looked into the recent commits and saw that you are now including MYMETA.* files in the tarball.
They should not be included there as they will be generated during the installation process. Same as for example the Makefile should not be in the tarball.

turnstep:
Ugh, thanks. Trying to do too many things at once!

oalders:
https://metacpan.org/pod/ExtUtils::MakeMaker#NO_MYMETA `NO_MYMETA` defaults to `false` anyway, so I think it's less confusing if it's just removed entirely. That seems simpler to me than having to figure out the double negative.

turnstep:
Works for me! MYMETA is now gone

oalders:
Thanks @turnstep! Closed via https://github.com/bucardo/dbdpg/commit/f316f892d99d8b9c868158b6509ac4b42b296e68

PR #115, opened 2023-04-05
Add GitHub Actions

szabgab:
To run the tests on every push and on every pull-request.

### DIFF
--- a/.github/workflows/with_pg.yaml
+++ b/.github/workflows/with_pg.yaml
@@ -1,0 +1,72 @@
+name: CI in Docker container
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


leonardoadame (approved):

turnstep:
Can you explain more exactly what this does and any downsides from doing so?

turnstep:
Is the specific versions something we would need to update here over time? For example, Postgres 11 is almost EOL so would think we try perhaps 14 and 15 for a while?

szabgab:
once you merge this pr you can update the list as you wish.

PR #129, opened 2024-09-27
Support binding native boolean false on 5.36 and newer

### LINKED ISSUE #125: v5.40's builtin false keyword cannot be passed in a bind parameter as the value of a boolean field
This seems like it's a bad thing, because writing `use v5.40;` on a script doesn't let you override the `true` & `false` builtins afterwards by using something like `use Mojo::JSON qw/ true false /;`, nor with `no builtin qw/ true false /`, so we're all stuck with v5.40's builtin boolean values, and maybe you should consider this.

DBD::Pg thinks `false` is an empty string, and PostgreSQL returns an error:

My example script:
```perl
#!/usr/bin/env perl

use v5.40;
use FindBin '$RealBin';
use lib "$RealBin/local/lib/perl5";

use DBI;

my $dbh = DBI->connect('dbi:Pg:dbname=dbname;host=127.0.0.1', 'user', 'pass', {AutoCommit => 1});

$dbh->do('INSERT INTO "foo" ("value") VALUES (?)', undef, false);
```

Output:
```
DBD::Pg::db do failed: ERROR:  invalid input syntax for type boolean: ""
CONTEXT:  unnamed portal parameter $1 = '' at ./true_false.pl line 11.
```

ilmari:
Fixes #125


esabol (approved):
Looks good to me! Thanks, @ilmari !

ilmari:
I just had a thought (and pushed a commit to do it, for discussion): should we respect `pg_bool_tf` and bind native booleans as `t`/`f` instead of `1`/`0` when it's set?

PR #134, opened 2024-11-07
Add example for using RETURNING in "insert into foo (...) returning id"

Tux:
Add example for using RETURNING in "insert into foo (...) returning id"


esabol (approved):

jonjensen:
Thanks! Good, plentiful examples in the documentation really help.

PR #140, opened 2025-05-14
08async fixes

rweikusat:
This fixes three minor mistakes in t/08async.t I found while testing my async-prepare changes:

Test #20 says "Method pg_ready() works after a non-async query", should be async as an async query was started.
Test #23 is a duplicate of test #21.
As test #23 is a duplicate of test #21, there are only 66 tests in this file and not 67.


esabol (approved):

esabol (approved):
Looks good to me!

PR #142, opened 2025-05-23
async prepare changes

rweikusat:
Transparently use PQsendPrepare instead of PQprepare for async queries to avoid blocking the client while waiting for a reply to the prepare request sent to the server.


esabol:
@rweikusat : Any comment regarding the possible regression mentioned in issue #183 ? Thanks!

rweikusat:
The obvious comment would be: The count isn't incremented because the statement hasn't yet been prepared. It will be incremented from `pg_db_ready` once the prepare completed. This matches the synchronous semantics of only increasing the count if the prepare actually worked. It's possible to trick that into an error by never waiting for the async prepare to complete and trigger another async prepare via execute while the old one is silently being waited for via `handle_old_async`. 

Code which can trigger this:

```
{
    my ($sth0, $sth1);

    $t=q{Silently waiting for an async prepare via handle_old_async works};
    $sth0 = $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => 5, pg_prepare_now => 1 });
    sleep(1);
    $$dbh{pg_switch_prepared} = 1;
    $sth1 = $dbh->prepare('select * from dbd_pg_test5 where t = ?', { pg_async => 5});
    $sth1->execute(2);
    eval {
        $dbh->pg_result();
    };
    is ($@, q{}, $t);
}

```

The `sleep(1)` is supposed to make sure that the async prepare completed on the server side. A second query is then created and fed into `->execute()` in a way which will cause another async prepare. The code which handles the execute will call `handle_old_async` to get rid of the running async prepare but  this won't increment the `imp_dbh->prepare_number` and hence, will use the same name twice.

My idea of a proper solution for this would be

```
diff --git a/dbdimp.c b/dbdimp.c
index a8dec1e..fba5a72 100644
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2547,9 +2547,10 @@ static int pg_st_prepare_statement (pTHX_ SV * sth, imp_sth_t * imp_sth)
         TRACE_PQSENDPREPARE;
         status = PQsendPrepare(imp_dbh->conn, imp_sth->prepare_name, statement, params,
                                imp_sth->PQoids);
-        if (status)
+        if (status) {
             imp_sth->async_status = STH_ASYNC_PREPARE;
-        else {
+            imp_dbh->async_sth = imp_sth;
+        } else {
             status = PGRES_FATAL_ERROR;
             _fatal_sqlstate(aTHX_ imp_dbh);
         }
@@ -5868,6 +5869,8 @@ static int handle_old_async(pTHX_ SV * handle, imp_dbh_t * imp_dbh, const int as
            to be sent & the result dealt with */
         if (async_sth && async_sth->async_status == STH_ASYNC_PREPARE
             && status == PGRES_COMMAND_OK) {
+            ++imp_dbh->prepare_number;
+
             ret = pq_send_prepared_query(aTHX_ imp_dbh, async_sth);
             if (!ret) {
                 TRACE_PQERRORMESSAGE;
```

This sets `imp_dbh->async_sth` after an async prepare was started from `pg_st_prepare_statement` (instead of only from `dbd_st_execute`) and then increments the prepare number from `handle_old_async` if the prepare actually succeeded.

Whether or not that's a use case worthy of being supported would be a different question. I can create a pull request for this from my repository.

esabol:
@rweikusat : Thank you for your response! I think that, in a multithreaded process, it does seem likely to be encountered?

I will test your variation of the fix later today. You might want to refer to the latest reproducing test script in  https://github.com/bucardo/dbdpg/issues/183#issuecomment-4260275430.

rweikusat:
This test script fails with

```
DBD::Pg::st pg_result failed: No asynchronous query is running at t/t40coll.t line 53.
# Looks like your test exited with 255 just after 1.
t/t40coll.t .... 

```

both for the original code and for the version I modified to prevent the name collision.  It also cannot possibly trigger the issue because it only ever actually prepares one statement.  

Inserting `$$dbh{pg_switch_prepared} = 1;`  before the loop to request that `execute` prepares a statement on the first execution instead of on the second causes the `sth_select2` statement to be prepared as well. If this is done, the original code triggers a name collision while my modified code doesn't anymore.

But I suggest to use the the minimal test case I posted in the previous comment instead. The code can be copied into t/08async.t immediately before the `$dbh->do('DROP TABLE dbd_pg_test5');` at the end and the plan number (line 21) increased to 70. This will cause a name collision with the unmodified code and won't anymore with the other. Both the test case and the fixed code are in the `handle-old-async-async-prepare` branch in my forked repository. The 'active' part of the fix can be disabled by commenting out line 5872 of dbdimp.c. A suitable command for testing this is

`make test TEST_FILES="t/01connect.t t/08async.t t/99cleanup.t"`

NB: The patch proposed in the other thread doesn't really fix the problem - improper handling of async prepare by `handle_old_async` plus `pg_st_prepare_statement` not setting `imp_dbh->async_sth` when it should. It just sidesteps it by increasing the `prepare_number` counter regardless of the outcome of the actual prepare.

rweikusat:
This fix doesn't work anymore for the current dbdimp.c because the "old query result auto-retrieve" code added to handle_old_async as effectively sidelines the code supposed to handle a completed async prepare below it. I'm working on merging the two.

esabol:
> This test script fails with
> 
> ```
> DBD::Pg::st pg_result failed: No asynchronous query is running at t/t40coll.t line 53.
> # Looks like your test exited with 255 just after 1.
> t/t40coll.t .... 
> ```
> 
> both for the original code and for the version I modified to prevent the name collision. It also cannot possibly trigger the issue because it only ever actually prepares one statement.

I don't think we're looking at the same test script? The one I linked to definitely prepares multiple different statements. I see 3 prepare_cached() and one do().

rweikusat:
I'm sorry but you just don't know what the code is actually doing. By
default (unless pg_prepare_now is passed), no prepare call ever
does a server-side prepare of anythinģ. This is done heuristically by
execute if the query meets certain criteria and is at least executed
$$dbh{pg_switch_prepared} times (default 2). This means insert gets
prepared on the second execution but the select doesn't as it runs only
once.

This will be printed in excruciating detail when DBI tracing is enabled
eg, by invoking DBI->trace(15).

PR #145, opened 2025-06-06
Proposed cancel changes.

rweikusat:
Splits `pg_db_cancel ` into a sychronous cancel operation callled `pg_db_cancel` and a function which just sends a cancel request to the server without waiting for the result called `pg_db_send_cancel`. Fixes the cancel result handling code in `pg_db_cancel` by invoking `pg_db_result` to wait for the result of the possibly cancelled query instead of partially duplicating the result handling code in `pg_db_cancel` itself.

Introduces a new internal function called `do_send_cancel` which contains just the cancel request sending logic and calls that from all places in the original code which used to duplicate this functionality.

Changes `pg_db_result` such that it no longer treats cancelled queries as fatal errors but returns zero and a SQLSTATE of 57014 instead.

Modifies t/08async.t to enable AutoCommit so that cancelled queries no longer cause transaction aborts which would need to be dealt with by rollbacks. Removes a now redundant `$dbh->commit()`. Changes `handle_old_async` cancel path to only issue a rollback if actually in a transaction to avoid warnings about issueing a rollback outside of a transaction.

Unrelated fix: Uses the correct TRACE macro (`TRACE_PQCANCEL`) when about to call `PQcancel` instead of the `TRACE_PQGETCANCEL` which was used in this place instead.

_This passes the test suite plus the two tests I added for_ `pg_db_send_cancel`.

### DIFF
--- a/Pg.pm
+++ b/Pg.pm
@@ -166,6 +166,7 @@ use 5.008001;
             DBD::Pg::db->install_method('pg_result'); ## NOT duplicated below!
             DBD::Pg::db->install_method('pg_rollback_to');
             DBD::Pg::db->install_method('pg_savepoint');
+            DBD::Pg::db->install_method('pg_send_cancel');
             DBD::Pg::db->install_method('pg_server_trace');
             DBD::Pg::db->install_method('pg_server_untrace');
             DBD::Pg::db->install_method('pg_type_info');
--- a/Pg.pm
+++ b/Pg.pm
@@ -4209,6 +4209,14 @@ that would have been returned by the asynchronous L</do> or L</execute> if it ha
 
   $result = $dbh->pg_result;
 
+=item B<pg_send_cancel>
+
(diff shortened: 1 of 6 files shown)


esabol:
@turnstep @rweikusat : Merging this PR caused the CI workflow to fail in `t/00basic.t` with `Error:  Can't load 'blib/arch/auto/DBD/Pg/Pg.so' for module DBD::Pg: blib/arch/auto/DBD/Pg/Pg.so: undefined symbol: pg_db_cancel at /usr/local/lib/perl5/5.40.3/x86_64-linux-gnu/DynaLoader.pm line 206.` Please investigate.

https://github.com/bucardo/dbdpg/actions/runs/20107180081/job/57694345958

rweikusat:
I think I have this fixed in the fix-pg_db_cancel-missing branch of my forked repository (plus another minor issue). I'll see if I can create a pull request from that (should work).

PR #148, opened 2025-08-19
C23 compliance.

rweikusat:
C23 eliminates the historical K&R syntax `void function()` to declare a function taking any number of arguments, making `void function()` equivalent to `void function(void)`. Because of this, types.h can no longer be compiled with a C23 compilers.

I've changed the function pointers to include prototypes and the quote/ dequote functions accordingly. I've updated the Perl code in types.c despite it doesn't seem to work anymore. While at that, I've also removed the unused estring argument from the dequote functions as the dbdimp.c code doesn't pass that, anyway.

### DIFF
--- a/quote.c
+++ b/quote.c
@@ -183,13 +183,13 @@ char * quote_circle(pTHX_ const char *string, STRLEN len, STRLEN *retlen, int es
 }
 
 
-char * quote_bytea(pTHX_ char *string, STRLEN len, STRLEN *retlen, int estring)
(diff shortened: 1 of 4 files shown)


esabol (approved):

nrdvana:
The gist of CPAN::InGit is that you run your own CPAN, point your Dockerfile at it, and then patch modules in it however you like.  It lets you move forward with new Docker images without waiting for upstream CPAN to catch up.

Reaster0:
ha indeed that could be interesting, I'll might use it in another contexts, also you the maintainer i guess so

PR #152, opened 2025-12-11
Fix pg db cancel missing

rweikusat:
This fixes a probably accidental editing error which cause the pg_db_send_cancel and pg_db_cancel to be "garbled together" while losing parts of the code (issue noticed by Ed Sabol).

It also changes the cancel path in handle_old_async to only do a rollback if there's actually a transaction running because I noticed a warning on make test and wanted to investigate that to ensure that it's harmless.

This doesn't address the other issue noted by Ed Sabol ("failed to create dbh ...") yet.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -5821,10 +5821,13 @@ static int handle_old_async(pTHX_ SV * handle, imp_dbh_t * imp_dbh, const int as
                 async_sth->prepare_name = NULL;
             }
 
-            /* We need to rollback! - reprepare!? */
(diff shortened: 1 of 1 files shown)


rweikusat:
Well, I have the time as I'm unemployed since my employer collapsed in January. And I actually enjoy working with code. That's at least hopefully more useful than doing job applications which - at best - get boilerplate "HR deleted this from the system" rejections.

:-)

rweikusat:
Thanks. I'll have a look at that.

PR #153, opened 2025-12-15
Fix async connect test

rweikusat:
Fix async connect test failing in GitHub action because of not using a password if one was provided. 

Based on an auto-generated fix by Toby Oetiker.

### DIFF
--- a/t/03dbmethod.t
+++ b/t/03dbmethod.t
@@ -2540,7 +2540,7 @@ ASYNC_CONNECT: {
     my ($dsn, $user) = get_test_settings();
     my ($rc);
 
-    sub dbi_connect
+    sub test_connect
     {
         return DBI->connect($dsn, $user, $ENV{DBI_PASS},
                             {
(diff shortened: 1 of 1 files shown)


rweikusat:
I have no idea what's going on there save that it happens during running Makefile.PL and that's not something I ever touched. I can principally fix this, but it's an unrelated issue.

rweikusat:
The problem is that `perl Makefile.PL` fails (can be seen by adding `-v` to the `cpanm` calls) because the MakeMaker version shipped with Perl 5.30 is too old to satisfy the

`use ExtUtils::MakeMaker 7.64;`

at the beginning of Makefile.PL.  A minimal fix is to add

` "ExtUtils::MakeMaker" : "7.64",`

to the `configure_requires` object/ dictionary in `META.json` which causes `cpanm --installdeps` to update MakeMaker if needed. `META.yml` should probably be changed as well. If that's ok, I can do a pull request for that once I've "unmessed" my upstream tracking branch.

PR #155, opened 2026-01-02
added ExtUtils::MakeMaker 7.64 to configure_requires

rweikusat:
Added a configure_requires depenency on ExtUtils::MakeMaker 7.64 to META.json and META.yml so that `perl Makefile.PL` again works with Perl 5.30 because cpanm will now update MakeMaker if the version included with a particular perl version is too old.

### DIFF
--- a/META.json
+++ b/META.json
@@ -47,6 +47,7 @@
    ],
    "configure_requires" : {
       "DBI" : "1.614",
+       "ExtUtils::MakeMaker" : "7.64",
       "version" : "0"
    },
    "license" : "perl",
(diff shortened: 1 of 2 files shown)


esabol:
@rweikusat : Can you please rebase this PR to see if all the CI tests all pass now? (See https://github.com/bucardo/dbdpg/pull/156#issuecomment-3874589634 for reference.)

META.json
```diff
@@ -47,6 +47,7 @@
    ],
    "configure_requires" : {
       "DBI" : "1.614",
+       "ExtUtils::MakeMaker" : "7.64",
```
esabol: Nitpick: The other lines use tabs to indent. This line you added uses spaces to indent. Change it to use a tab instead for the sake of consistency?

PR #156, opened 2026-01-02
updated copyright end years to stop 99_lint.t tests from failing

rweikusat:
Since I noticed that, I thought I could just fix it quickly.


turnstep:
Sorry, did not see this until now, but already handled with 0cd31e951392eac445eda68eebba737d60e872e2

PR #158, opened 2026-02-09
pg_error_field(): don't modify argument string (#157)

### LINKED ISSUE #157: pg_error_field rudely modifies its argument
The `pg_error_field` method breaks encapsulation and directly modifies the character buffer of its argument, even if that argument is a (nominally read-only) constant or a (nominally copy-on-write) shared string.

For example:

```perl
my $field_name = 'severity';
$dbh->pg_error_field($field_name);
say "oops: $field_name";  # oops: SEVERITY
```

There is no apparent modification of `$field_name`, but after the call the string is all uppercase.

mauke:
Fixes #157.

### DIFF
--- a/Pg.xs
+++ b/Pg.xs
@@ -447,7 +447,10 @@ pg_error_field(dbh, fieldname)
     SV * dbh
     char * fieldname;
     CODE:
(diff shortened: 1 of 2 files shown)


esabol (approved):
LGTM!

Issue #159, opened 2026-02-25
github CI tests and ownership issues


turnstep:
Some of the tests are now showing this warning:

```
fatal: detected dubious ownership in repository at '/__w/dbdpg/dbdpg'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/dbdpg/dbdpg
```

I think this might be because t/99_lint.t now does a "git ls-files".

I'm not clear on what the solution would be for this, as I am not very familiar with github workflows. Ideas?

turnstep:
I think we are good now, thanks for the fixes.

PR #160, opened 2026-02-25
Add Dependabot configuration

esabol:
As discussed in issue #159, this should keep the dependencies of the GitHub Actions workflows up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep the GitHub Actions used in .github/workflows/* up to date
+# Reference: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"                  # Root of the repository
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10    # Prevent too many open PRs
+    # Group updates together (newer Dependabot feature)
+    groups:
+      actions:
+        dependency-type: "production"


esabol:
So Dependabot is reporting errors... Go to **Insights > Dependency graph > Dependabot** and you will see  this:
Screen Shot 2026-03-15 at 5 07 27 AM

Google says:
> This specific error occurs because your repository's Branch Protection Rules are set to a "Linear History" requirement, which forbids the use of merge commits.
> When Dependabot attempts to create or update a PR, it may default to a merge strategy that includes a merge commit, or it may be trying to merge its changes into your base branch using a method that your settings currently block.
> To resolve this, you have two main options:
>
> **Option 1: Adjust Branch Protection Settings**
> The most common fix is to allow the specific merge types that Dependabot uses.
> 1. Go to your repository Settings > Branches.
> 2. Find the Branch protection rules for your default branch (usually main or master) and click Edit.
> 3. Look for the setting Require linear history (which "Prevents merge commits from being pushed to matching branches").
> 4. Uncheck this box if you are comfortable allowing merge commits. This is the quickest way to let Dependabot finish its work.
>
> **Option 2: Change Repository Merge Defaults**
> If you want to keep your linear history requirement, ensure your repository is set up to handle PRs via "Squash" or "Rebase" only:
> 1. Go to Settings > General.
> 2. Scroll down to the Pull Requests section.
> 3. Ensure Allow squash merging or Allow rebase merging is enabled.
> 4. Uncheck "Allow merge commits" globally for the repository. Dependabot will then attempt to use one of the other allowed methods. 

@turnstep: I'm guessing you prefer Option 2?

esabol:
> I do, just to keep the git history as clean as possible. Not a hill I'm going to die on, however, if you want to make a strong case for Option 1. :)

I think this is the only GitHub repo I've come across that has "Require linear history" enabled, so it's a bit unusual, but it doesn't really matter to me. Can you make the configuration change in Option 2?

PR #163, opened 2026-03-14
Redux: Fix hangup of execution of queries like "COPY TO STDOUT" in asynchronous mode (now with tests)

esabol:
This is an updated version of @true-alex's pull request #101. It is meant to address issue #98.

`$dbh->do("COPY (select * from unnest(ARRAY[1,2,3,4,5,6,7,8]::int[]) as mytable(a)) TO STDOUT", {pg_async => PG_ASYNC});` still hangs with DBD::Pg 3.19.0.

I have applied @true-alex's proposed fix to the latest master branch and added tests to `t/08async.t`.

The proposed fix indeed prevents hanging the process, but I don't think the return value of `$dbh->pg_result()` is correct. Once `$dbh->pg_ready` becomes true after executing `$dbh->do("COPY (select * from unnest(ARRAY[1,2,3,4,5,6,7,8]::int[]) as mytable(a)) TO STDOUT", {pg_async => PG_ASYNC});`, `$dbh->pg_result()` currently returns -1, but I think it should return 0 (the number of rows affected). Am I right about that, @turnstep ?

Marking PR as a draft for now.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -5589,6 +5589,9 @@ long pg_db_result (SV *h, imp_dbh_t *imp_dbh)
             imp_dbh->last_result = result;
             imp_dbh->result_clearable = DBDPG_TRUE;
         }
+        if (rows == -1) {
+            break;
+        }
     }
 
     if (imp_sth && imp_sth == imp_dbh->async_sth) {
--- a/t/08async.t
+++ b/t/08async.t
@@ -26,7 +26,7 @@ if (! $dbh_noerr) {
 $dbh_noerr->{RaiseError} = 0;
 $dbh_noerr->{PrintError} = 0;
 
-plan tests => 117;
+plan tests => 128;
 
 isnt ($dbh, undef, 'Connect to database for async testing');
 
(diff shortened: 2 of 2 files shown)


jjn1056:
Chiming in on the return value question. After tracing through the code, I think `-1` is correct and consistent. Here's why:

**Synchronous COPY already returns `-1`**

When you run COPY synchronously, `dbd_st_execute` hits the `PGRES_COPY_OUT/IN/BOTH` case and returns `-1` ([dbdimp.c line 3897](https://github.com/bucardo/dbdpg/blob/master/dbdimp.c#L3897)). DBI's `do()` passes that through:

```perl
# Synchronous — returns -1
my $ret = $dbh->do("COPY mytable TO STDOUT");
# $ret is -1
```

If `pg_result()` returned `0` for async COPY, the two paths would give different answers for the same operation:

```perl
# Async — would return 0 (inconsistent)
$dbh->do("COPY mytable TO STDOUT", {pg_async => PG_ASYNC});
# ... pg_ready() ...
my $ret = $dbh->pg_result();
# $ret would be 0, but sync do() returns -1 for the same statement?
```

**`-1` has the right DBI semantics**

DBI defines `-1` as "number of rows is unknown or not available." That's precisely the situation — a COPY statement starts a data transfer, it doesn't affect rows. The actual row count only becomes known after draining with `pg_getcopydata`. Returning `0` would mean "zero rows affected," which isn't accurate either.

**The code is already consistent**

`pg_db_result` already sets `rows = -1` for COPY results in its switch statement ([dbdimp.c line 5735](https://github.com/bucardo/dbdpg/blob/master/dbdimp.c#L5735)). The 3-line fix in this PR correctly lets that value escape the `PQgetResult` loop instead of hanging. No additional changes needed.

**Suggested test fix**

The TODO test should just assert `-1`:

```perl
is ($res, -1, "pg_result() after async COPY TO STDOUT returns -1 (rows unknown during COPY)");
```

esabol:
Thanks for your thoughts on this, @jjn1056. I found your points persuasive, especially this:

> DBI defines -1 as "number of rows is unknown or not available." That's precisely the situation — a COPY statement starts a data transfer, it doesn't affect rows. The actual row count only becomes known after draining with pg_getcopydata. Returning 0 would mean "zero rows affected," which isn't accurate either.

What do you think, @turnstep ?

turnstep:
Yeah, that makes sense now that we see it all in context.

esabol:
OK, I have rebased and squashed commits. CI pass is clean. Ready for review.

@turnstep : Let me know if you want any changes to the tests.

esabol:
I have rebased and resolved merge conflicts on this PR. All tests still pass. Merging before there are new conflicts. :smile:

PR #164, opened 2026-03-14
Fix test failures found by CPAN Testers

esabol:
Some tests by various cpantesters exhibit the following message:

`Use of uninitialized value $dbh in concatenation (.) or string at t/01connect.t line 25.`

when the machine running the test cannot connect to a PostgreSQL server.

Example: https://matrix.perl-magpie.org/results/47814d4c-1fcc-11f1-9b05-1d0d6e8775ea

This PR should fix that.

### DIFF
--- a/t/01connect.t
+++ b/t/01connect.t
@@ -117,7 +117,7 @@ $t=q{Calling DBI->connect() with proper quoting but bad port gives expected erro
 ## connections on Unix domain socket "/tmp/.s.PGSQL.1"?
 $bad_dsn = q{dbi:Pg:dbname='dbdpg \'spacey\' name';port=1};
 eval { DBI->connect($bad_dsn, '', '', {RaiseError=>1}) };
-like ($@, qr/DBI.*\Q.s.PGSQL.1\E\b/s, $t);
+like ($@, ($^O =~ /Win/ ? qr/DBI/s : qr/DBI.*\Q.s.PGSQL.1\E\b/s), $t);
 
  SKIP: {
      if ($pglibversion <  100000) {


esabol:
All Windows cpantesters are failing this test in `t/01connect.t`:
```Perl
$t=q{Calling DBI->connect() with proper quoting but bad port gives expected error};
## An otherwise correct call but to an invalid port gives a message like this:
## DBI connect('dbname='dbdpg \'spacey\' name';port=1','',...) failed: 
## could not connect to server: No such file or directory
## Is the server running locally and accepting
## connections on Unix domain socket "/tmp/.s.PGSQL.1"?
$bad_dsn = q{dbi:Pg:dbname='dbdpg \'spacey\' name';port=1};
eval { DBI->connect($bad_dsn, '', '', {RaiseError=>1}) };
like ($@, qr/DBI.*\Q.s.PGSQL.1\E\b/s, $t);
```
It seems Windows returns a different error message:
```
#   Failed test 'Calling DBI->connect() with proper quoting but bad port gives expected error'
#   at t\01connect.t line 120.
#                   'DBI connect('dbname='dbdpg \'spacey\' name';port=1','',...) failed: could not connect to server: Connection refused (0x0000274D/10061)
# 	Is the server running on host "localhost" (::1) and accepting
# 	TCP/IP connections on port 1?
# could not connect to server: Connection refused (0x0000274D/10061)
# 	Is the server running on host "localhost" (127.0.0.1) and accepting
# 	TCP/IP connections on port 1? at t\01connect.t line 119.
# '
#     doesn't match '(?^s:DBI.*\.s\.PGSQL\.1\b)'
```
Perhaps this could be fixed by changing the test like so?
```Perl
like ($@, ($^O =~ /Win/ ? qr/Connection refused/s : qr/DBI.*\Q.s.PGSQL.1\E\b/s), $t);
```

PR #176, opened 2026-03-18
Add non-blocking async COPY FROM support (pg_putcopydata_async, pg_putcopyend_async, pg_flush)

jjn1056:
Closes #177

## Summary

Adds three new methods to enable non-blocking COPY FROM STDIN for async Perl libraries:

- **pg_putcopydata_async** — non-blocking version of pg_putcopydata. Returns 1/0/-1 matching PQputCopyData semantics.
- **pg_putcopyend_async** — non-blocking version of pg_putcopyend. Polls for server result without blocking.
- **pg_flush** — exposes PQflush for manual output buffer management. Returns 0/1/-1 matching PQflush semantics.

## Motivation

DBD::Pg has async support for queries (pg_async/PG_ASYNC) and async COPY TO (pg_getcopydata_async), but COPY FROM is unconditionally blocking. pg_putcopydata blocks inside PQputCopyData when the TCP send buffer fills because DBD::Pg never  calls PQsetnonblocking.

This prevents async Perl libraries (Future::IO-based, EV::Pg-style, IO::Async) from performing non-blocking bulk data loading. COPY is the standard PostgreSQL mechanism for bulk import and is dramatically faster than multi-row INSERT.

## Approach

The implementation follows the pattern already established by pg_getcopydata/pg_getcopydata_async:

- pg_db_putcopydata gains an `async` flag parameter (0 for blocking, 1 for async)
- The blocking path (async=0) is unchanged — existing pg_putcopydata behavior is identical
- The async path calls PQsetnonblocking(conn, 1) on first use, scoped to COPY state (safe because no other operations are permitted during COPY)
- PQsetnonblocking is restored to 0 when pg_putcopyend_async completes
- Return values match libpq conventions exactly (verified against EV::Pg and libpq docs)

## Return value conventions

pg_putcopydata_async:  1 = queued, 0 = buffer full (retry), -1 = error
pg_putcopyend_async:   1 = done,   0 = not ready (poll),    -1 = error
pg_flush:              0 = flushed, 1 = pending (poll),     -1 = error

After pg_putcopydata_async returns 1, the caller calls pg_flush to push data to the server. This matches the standard libpq pattern: PQputCopyData queues, PQflush sends.

## Implementation details

- **Pg.h**: Added TRACE_PQFLUSH and TRACE_PQSETNONBLOCKING macros
- **dbdimp.h**: Added `copy_nonblocking` field to imp_dbh_t, new function declarations
- **dbdimp.c**: Modified pg_db_putcopydata to accept async flag, added pg_db_putcopyend_async (uses copystate=-1 sentinel for "end sent, awaiting result" phase), added pg_db_flush
- **Pg.xs**: Added XS bindings following the getcopydata/getcopydata_async pattern
- **Pg.pm**: Added install_method calls and POD documentation with examples

## Tests

27 new test assertions in t/07copy.t covering:
- Basic async put/flush/end cycle with data verification
- pg_flush return values
- Wrong-state errors (not in COPY, COPY OUT state, no argument)
- do() rejection during async COPY IN
- Recovery after interrupted COPY
- Binary COPY round-trip via async methods
- Multiple async COPY cycles on the same connection (PQsetnonblocking toggling)
- Large data set (1000 rows) with buffer-full handling
- Backward compatibility (blocking methods still work after async used)

Full test suite passes: 3,773 tests, 0 failures, 0 regressions.

## Backward compatibility

Zero risk. The blocking pg_putcopydata is unchanged — the XS wrapper passes async=0. No existing behavior is affected. New methods are purely additive.


jjn1056:
This PR does not supersede PR #163. They address different directions of COPY:

- **#176** (this PR): Adds non-blocking async **COPY FROM STDIN** (write/import side) — `pg_putcopydata_async`, `pg_putcopyend_async`, `pg_flush`
- **#163**: Fixes a hang when executing **COPY TO STDOUT** in async mode (read/export side) via `pg_async => PG_ASYNC`

They're complementary and could both land independently.

Re: spelling and perlcritic failures — working on fixes now.

t/07copy.t
```diff
@@ -426,6 +426,263 @@ is $@, '', 'pg_putcopydata in binary mode works'
...
+};
+is ($@, q{}, $t);
+is ($result, 1, $t);
+$dbh->pg_flush();
+$dbh->pg_putcopyend();
+$dbh->commit();
+
+# Binary COPY with async methods
+
+$dbh->do('CREATE TEMP TABLE binarycopy_async AS SELECT 1::INTEGER AS x');
```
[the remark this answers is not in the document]
jjn1056: agreed, I should have followed your conventions more closely.  done.

t/07copy.t
```diff
@@ -426,6 +426,263 @@ is $@, '', 'pg_putcopydata in binary mode works'
 $t=q{COPY in binary mode round trips};
 is_deeply ($dbh->selectall_arrayref('SELECT * FROM binarycopy'), [[1],[1]], $t); ## nospellcheck
 
+##
+## Test the async COPY methods
+##
+
+my $async_table = 'dbd_pg_test_async_copy';
+$dbh->do(qq{CREATE TABLE $async_table(id integer, name text)});
```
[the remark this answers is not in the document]
jjn1056: got it, followed that.

jjn1056:
@esabol Thanks for looking at this.  Sorry took me so long to get back to you; I got pulled into a critical end of quarter thing at the office and was tapped out :).

I went thru all these and doubled checked for regressions, etc.   I think this along with PR 163 gets us really stellar asynchronous support for COPY, which is for my use cases a big win.

Let me know if you spot anything else I need to tweak, and again thanks for your time running this all important project for the Perl community!

esabol:
Thanks, @jjn1056 ! The latest commit passes the CI workflow. Everything looks good to me, but we need to get @turnstep to do the final review.

esabol (approved):

jjn1056:
One of the parts I wasn't sure about was if we needed to version gateway any of this.   As far as I could tell this should work even on vastly out of date versions of Postgresql, like 9.5 but I might try to spin some sort dockers with that just to double check.

esabol:
@jjn1056 wrote:
> One of the parts I wasn't sure about was if we needed to version gateway any of this. As far as I could tell this should work even on vastly out of date versions of Postgresql, like 9.5 but I might try to spin some sort dockers with that just to double check.

Ah. Thanks for checking. Technically, according to the README, DBD::Pg is supposed to support PostgreSQL going all the way back to 8.0. It's unfortunate we don't have the CI apparatus to test that.

jjn1056:
I'll see if I can find anyone with a working postgresql 8 docker.   It would be handy because I want to look at pipelining next and that requires Pg 14+ so we will need to think about some sort of conditional compile eventually; so if we need it now would be good to get agreement on the way we should do that.

dbdimp.c
```diff
@@ -4615,6 +4644,186 @@ int pg_db_putcopyend (SV * dbh)
...
+   Sends the COPY end marker and attempts to collect the server result
+   without blocking. Restores blocking mode on completion.
+
+   This function is designed to be called repeatedly in a poll loop:
+   - First call: sends PQputCopyEnd, attempts flush and result check
+   - Subsequent calls (when returning 0): polls for the server result
+
+   Returns:
+    1 = COPY completed successfully, connection is back in normal mode
+    0 = server result not ready yet (caller should poll socket for read-ready)
```
turnstep: I like read-ready better than read-readiness. Let's standardize on the former.

Pg.pm
```diff
@@ -4433,12 +4436,106 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+normal queries can be issued.
+
+=head3 B<pg_flush>
+
+Flushes the libpq output buffer. Wraps C<PQflush> directly. Used after
+L</pg_putcopydata_async> returns 1 to push queued data to the server.
+
+Return values match C<PQflush>:
+
+  0  = all data flushed successfully
```
turnstep, line 4545: Do we really need to match C return values here? Seems a little off that the other async methods return 0 on a problem, and 1 on success, but this is reversed. Or maybe change the others? Just a random thought.

Pg.pm
```diff
@@ -4433,12 +4436,106 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+server. If C<pg_flush> returns 1 (data pending), poll the socket for
+write-readiness and call C<pg_flush> again.
+
+Example usage:
+
+  ## Simple usage (flush after each row):
+  use IO::Select;
+  $dbh->do("COPY mytable(id, flavor, slices) FROM STDIN");
+  for my $row ("123\tPepperoni\t3\n", "314\tMushroom\t8\n") {
+      $dbh->pg_putcopydata_async($row);
```
turnstep: Example should be checking return value before going on

Pg.pm
```diff
@@ -4433,12 +4436,106 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+      while ($dbh->pg_flush()) {
```
turnstep: This matches 1 and -1, so we should be more explicit

Pg.pm
```diff
@@ -4433,12 +4436,106 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+          IO::Select->new($dbh->{pg_socket})->can_write();
```
turnstep: Not the best example to have it create a new object every iteration of the loop.

dbdimp.c
```diff
@@ -4528,23 +4541,34 @@ int pg_db_putcopydata (SV * dbh, SV * dataline)
...
+                return -1;
+            }
         }
+        /* For async mode, caller is responsible for calling pg_flush */
+        if (TEND_slow) TRC(DBILOGFP, "%sEnd pg_db_putcopydata (1)\n", THEADER_slow);
+        return 1;
     }
-    else if (0 == copystatus) { /* non-blocking mode only */
+    else if (0 == copystatus) {
+        /* Non-blocking mode: output buffer full, caller should wait for write-ready */
```
turnstep, line 4550: Should we set imp_dbh->copystate here?

Pg.pm
```diff
@@ -4433,12 +4436,106 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+      }
+      die "COPY error" if $status == -1;
+      while ($dbh->pg_flush()) {     # push to server
+          $sel->can_write();
+      }
+  }
+
+  ## Non-blocking end: poll until server confirms
+  while ((my $end = $dbh->pg_putcopyend_async()) == 0) {
+      $sel->can_read();
```
turnstep: Er...can_read or can_write? Can buffer be full here?

dbdimp.c
```diff
@@ -4505,20 +4505,33 @@ int pg_db_getcopydata (SV * dbh, SV * dataline, int async)
...
 
+    /* Enable non-blocking mode for async callers (safe during COPY: no other ops allowed) */
+    if (async && !imp_dbh->copy_nonblocking) {
+        TRACE_PQSETNONBLOCKING;
+        if (PQsetnonblocking(imp_dbh->conn, 1) != 0) {
+            _fatal_sqlstate(aTHX_ imp_dbh);
+            TRACE_PQERRORMESSAGE;
+            pg_error(aTHX_ dbh, PGRES_FATAL_ERROR, PQerrorMessage(imp_dbh->conn));
+            if (TEND_slow) TRC(DBILOGFP, "%sEnd pg_db_putcopydata (error: setnonblocking)\n", THEADER_slow);
+            return -1;
```
turnstep, line 4520: Do we need to change copystate?

t/07copy.t
```diff
@@ -426,6 +426,262 @@ is $@, '', 'pg_putcopydata in binary mode works'
...
+        last;
+    }
+    # If buffer full (0), poll and retry
+    while ($row_result == 0) {
+        select(undef, undef, undef, 0.001);
+        $row_result = $dbh->pg_putcopydata_async("$i\tRow number $i\n");
+    }
+    # Flush after each successful queue
+    my $flush = $dbh->pg_flush();
+    while ($flush == 1) {
```
turnstep, line 542: If flush has an error, we simply bypass this whole section? Should check for a -1

turnstep (commented):
Left a few notes, but I need to circle back when I can give this a lot more focus. Excellent work, though, I'm excited to get this in and have some solid async support finally.

dbdimp.c
jjn1056: Good catch — standardized on "write-ready" / "read-ready" throughout. The C code already used the short form; updated the three instances in Pg.pm to match.

Pg.pm
jjn1056: I thought about this and leaned toward keeping each function's return values matching its libpq counterpart — pg_flush matches PQflush, pg_putcopydata_async matches PQputCopyData, etc. Anyone coding against these will likely have the libpq docs open, so diverging from those conventions could cause more confusion than the internal inconsistency. That said, the existing pg_getcopydata_async already follows the same pattern (return values match PQgetCopyData), so there's precedent for this approach. Happy to change it if you feel strongly though.

Pg.pm
jjn1056: You're right — removed the simple example entirely. It had multiple issues (unchecked return value, `while ($dbh->pg_flush())` matching both 1 and -1, new IO::Select on every iteration) and trying to fix all of those makes it no longer "simple." The robust example that follows already demonstrates proper return value checking, explicit flush loop conditions, and reuses a single IO::Select object, so it's a better single reference for users.

Pg.pm
jjn1056: Addressed — removed the simple example (see reply on the previous comment).

Pg.pm
jjn1056: Addressed — removed the simple example (see reply on the previous comment).

dbdimp.c
jjn1056: I don't think we should change copystate here — we're still in COPY_IN, the buffer being full is a transient condition. The caller will retry the same `pg_putcopydata_async` call after the socket becomes write-ready, and that re-entry checks for `PGRES_COPY_IN` as a precondition. Changing copystate would break that check. What do you think?

turnstep:
> Ah. Thanks for checking. Technically, according to the README, DBD::Pg is supposed to support PostgreSQL going all the way back to 8.0. It's unfortunate we don't have the CI apparatus to test that.

Don't sweat this too much. Support for old unsupported versions of Postgres (and v8 is so damn old) has always been on a "best effort" basis. Maybe we should make that more explicit in the docs? I still spin up 8.4 for testing, but it's getting harder and harder to compile those old versions. I never want to stop supporting older versions just because they are old, and am willing to have the code go through some contortions to keep them working, but rest assured, there are limits. :)

Pg.pm
turnstep: Nah, it's fine as is. In an ideal world we would all be using the same system, but that ship sailed long, long ago.

dbdimp.c
turnstep: Going to have to circle back to this when I have a clear schedule and a clear head. It all made sense when I wrote that, but now I'm not sure. :)

Pg.pm
jjn1056: You're right — `can_read` is only correct for the result-polling phase, but `pg_putcopyend_async` also returns 0 during the flush phase (which needs write-ready). Rather than try to solve that with IO::Select, I replaced the example with `Time::HiRes::sleep`-based polling to match the style used by the existing async examples (`pg_ready`, `pg_cancel`, etc.) elsewhere in the POD. This also sidesteps the read vs write ambiguity entirely. In practice, users of these methods will be in a real event loop (IO::Async, Mojo::IOLoop, AnyEvent, etc.) where they'd register the socket for both read and write events — but examples for specific event loops probably belong in a cookbook or separate document rather than the core POD.

dbdimp.c
jjn1056: I don't think so — `copystate` tracks the COPY protocol state (COPY_IN, COPY_OUT, etc.) while `copy_nonblocking` tracks whether we've enabled non-blocking mode on the connection. Switching to non-blocking doesn't change the protocol state — we're still in COPY_IN. The two flags are intentionally orthogonal. What do you think?

t/07copy.t
jjn1056: Good catch — added -1 checks on `pg_flush` in the large data set test. Now it sets `$async_ok = 0` and breaks out of the loop on flush error, same pattern as the `pg_putcopydata_async` error check above it.

esabol: I'm wondering if lines 546-556 could be simplified to the following?
```perl
while ($flush == 1) {
    select(undef, undef, undef, 0.001);
    $flush = $dbh->pg_flush();
}
if ($flush == -1) {
    $async_ok = 0;
    last;
}
```
Less redundancy... Possibly more efficient.

jjn1056: Good catch — adopted your version. It's cleaner and also fixes a subtle bug: the previous `last` inside the `while` loop only exited the inner loop, so the outer `for` would keep going after a flush error. Your restructuring puts the error check after the `while`, so `last` correctly exits the `for` loop.

esabol:
@turnstep : Is there anything else to do here? If not, I suggest merging this PR, if it passes your review, and then releasing DBD::Pg 3.21. Or hold off on merging this PR and release DBD::Pg 3.21 without it?

esabol:
Unfortunately, we have a merge conflict now, a minor one with `.perlcriticrc`. @jjn1056, can you rebase and resolve the conflict?

esabol:
> Unfortunately, we have a merge conflict now, a minor one with `.perlcriticrc`. @jjn1056, can you rebase and resolve the conflict?

Never mind. I was able to resolve the merge conflict using GitHub easily enough.

jjn1056:
@esabol sorry I didn't see the comment, had a busy week with the boss on vacation :)

esabol:
Another merge conflict has crept in, this time with dbdimp.c. Please resolve when you have time, @jjn1056.

esabol:
> Another merge conflict has crept in, this time with dbdimp.c. Please resolve when you have time, @jjn1056.

It's been a couple weeks. Are you still with us, @jjn1056 ?

jjn1056:
@esabol hey got super busy on the job, let me take a look.

jjn1056:
@esabol should be good to go now, let me know if you run into issues.

esabol:
> @esabol should be good to go now, let me know if you run into issues.

Thank you, @jjn1056 ! Much appreciated. All tests pass. I hope we can get this merged soon.

@turnstep, what do you think?

turnstep:
I will try to give this a fresh look this week, but I don't foresee any problems.

Pg.pm
```diff
@@ -4462,12 +4465,88 @@ the COPY statement. Returns a 1 on successful input. Examples:
...
+completes.
+
+Note: the connection performing the COPY is restricted to COPY operations until
+the COPY ends. However, the non-blocking methods allow the event loop to service
+other connections and tasks between calls, which is the primary benefit over the
+blocking variants.
+
+Return values match C<PQputCopyData>:
+
+  1  = data queued successfully (caller should call L</pg_flush> to send)
```
turnstep, line 4483: Cannot use markup here as the indent makes it verbatim

Issue #181, opened 2026-03-30
Proposal: Add pipeline mode support (PostgreSQL 14+)


jjn1056:
Pipeline mode has been on the TODO list for a while (`Support pipeline mode: https://www.postgresql.org/docs/14/libpq-pipeline-mode.html`). We have a working implementation and would like to get feedback on the API design before pushing further.

## Motivation

Pipeline mode (libpq, PG14+) batches multiple queries into fewer network round trips. Instead of the normal send-wait-receive cycle per query, the client queues N queries and collects N results after a sync point. This can dramatically reduce latency for workloads that issue many independent queries — batch inserts, bulk prepared statement execution, etc.

Every major PostgreSQL driver has added or is adding pipeline support: psycopg3 (`conn.pipeline()`), Ruby pg (`enter_pipeline_mode`), pgx (`SendBatch`), tokio-postgres (implicit pipelining).

## Proposed API

Low-level wrappers matching the Ruby pg gem's approach — thin 1:1 mappings of the libpq pipeline functions. Users manage the lifecycle explicitly.

| Method | Wraps | Returns |
|--------|-------|---------|
| `pg_enter_pipeline_mode` | `PQenterPipelineMode` | 1 success, 0 fail |
| `pg_exit_pipeline_mode` | `PQexitPipelineMode` | 1 success, 0 fail |
| `pg_pipeline_status` | `PQpipelineStatus` | 0=off, 1=on, 2=aborted |
| `pg_pipeline_sync` | `PQpipelineSync` | 1 success, 0 fail |
| `pg_getresult` | `PQgetResult` | hashref or undef |
| `pg_send_query_params` | `PQsendQueryParams` | 1 success, 0 fail |
| `pg_send_prepare` | `PQsendPrepare` | 1 success, 0 fail |
| `pg_send_query_prepared` | `PQsendQueryPrepared` | 1 success, 0 fail |
| `pg_send_flush_request` | `PQsendFlushRequest` | 1 success, 0 fail |
| `pg_flush` | `PQflush` | 0=flushed, 1=pending, -1=error |

`pg_getresult` returns a hashref:

```perl
{
    status    => 2,              # ExecStatusType integer
    error     => undef,          # error message or undef
    ntuples   => 3,              # row count
    nfields   => 2,              # column count
    cmdtuples => '',             # affected rows (DML)
    rows      => [[...], ...],   # only present for TUPLES_OK
}
```

## Usage example

```perl
[...]

PR #182, opened 2026-03-30
Add pipeline mode support for PostgreSQL 14+ (RFC)

jjn1056:
Working reference implementation of the pipeline mode API proposed in #181. See the issue for API rationale, open questions, and design alternatives.

## Summary

- 10 new `$dbh` methods wrapping libpq's pipeline API
- Both `?` and `$1`/`$2` placeholder syntax supported via C-level converter
- Compile-time gating (`#if PGLIBVERSION >= 140000`) — builds cleanly against older libpq, methods croak if called
- COPY operations guarded against in pipeline mode
- 109 tests covering lifecycle, queries, errors, prepared statements, flush, placeholder edge cases
- Full author test suite passes (4,757 tests, `AUTHOR_TESTING=1 RELEASE_TESTING=1`)

This is an RFC to get a stake in the ground and start a conversation. The pipeline mode TODO item has been in the project for a while, and the public API here is a proposal. The code works and is well-tested, but we'd like maintainer input on the API design and approach before considering it final. See #181 for the full discussion.

## Test plan

- [x] Enter/exit lifecycle and idempotency
- [x] Pipeline sync with result collection
- [x] INSERT/SELECT pipelines with data verification
- [x] Error handling: FATAL_ERROR -> PIPELINE_ABORTED -> recovery
- [x] Prepared statements in pipeline (PREPARE + EXECUTE)
- [x] Flush request without sync point
- [x] COPY guards (putcopydata/putcopyend/getcopydata croak)
- [x] Zero-row SELECT, NULL params, aborted status, consecutive syncs
- [x] `?` placeholder conversion: basic, multiple, inside quotes/comments/dollar-quoting
- [x] `$1` passthrough when already present
- [x] Large pipeline (100 queries)
- [x] Full author test suite with AUTHOR_TESTING=1 RELEASE_TESTING=1

🤖 Generated with [Claude Code](https://claude.ai/code)


Changes
```diff
@@ -4,6 +4,8 @@ RT refers to rt.cpan.org
 
 Version 3.20.0  (released March 19, 2026)
 
+ - Add pipeline mode support for PostgreSQL 14+
```
esabol: This doesn't belong here. (3.20.0 has already been released.)

I think @turnstep usually takes care of updating the Changes file when a new release is being prepared, so maybe leave this out of the PR?

Pg.pm
```diff
@@ -4439,6 +4449,148 @@ When you are finished with pg_putcopydata, call pg_putcopyend to let the server
...
+
+Example:
+
+  $dbh->pg_enter_pipeline_mode();
+  $dbh->pg_send_query_params('INSERT INTO t VALUES ($1)', [1]);
+  $dbh->pg_send_query_params('SELECT * FROM t', []);
+  $dbh->pg_pipeline_sync();
+
+  my $ins = $dbh->pg_getresult();   # {status => 1}
+  $dbh->pg_getresult();              # undef (separator)
```
esabol, line 4544: Nitpick: The whitespace before the `#` is irregular and doesn't line up, which I think is the intent. Applies to all lines 4533-4538.

Changes
turnstep: I am pretty agnostic about this - it all gets cleaned up before the release, but anybody adding things to the Changes file (even if in the wrong spot) means less work for me, so +1.

esabol: Ok, but I still wouldn't put in under `Version 3.20.0  (released March 19, 2026)`. I suggest adding a `Version TBD (released TBD)` and putting it under that.

jjn1056: sorry about that, an oversight.   Removed :)

PR #184, opened 2026-04-17
Handle async prepare correctly in handle_old_async

rweikusat:
When `handle_old_async` is invoked in wait mode and encounters a successful async prepare, it won't increment `imp_dbh->prepare_number` and hence, the next server-side prepared statement which will be created afterwards will cause a name collision. 

This pull request really fixes three related issues:

- `pg_st_prepare_statement` doesn't set `imp_dbh->async_sth` after starting an async prepare
- the code to auto-retrieve results in `handle_old_async` incorrectly treats the result of a prior async prepare as query result despite the query wasn't executed yet
- the code supposed to deal with a successful async prepare in `handle_old_async` fails to increment `imp_dbh->prepare_number` and thus, causes the abovementioned name collision

Further, it does some minor cleanups to simplify the `handle_old_async` code:

- remove two redundant `asyncflag & PG_OLDQUERY_WAIT` tests from code which only runs if this condition was true
- use existing `async_sth` variable (initialized with `imp_dbh->async_sth`) instead of repeating `imp_sth->async_sth`

Lastly, it adds a test case for catching the name collision which will occur of the `++imp_dbh->prepare_number` in `handle_old_async` isn't there (aka commented out).

### DIFF
--- a/t/08async.t
+++ b/t/08async.t
@@ -398,7 +398,7 @@ $t=q{Database method pg_result returns correct result after execute};
 is ($res, 2, $t);
 
 {
-    my ($sth0, $sth1);
+    my ($sth0, $sth1, $old_switch);
 
     #
     # Test that handle_old_async deals correctly with async prepares.
@@ -411,6 +411,7 @@ is ($res, 2, $t);
     $sth0= $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => PG_ASYNC, pg_prepare_now => 1 });
 
     # Tell execute that it should prepare on first execution.
+    $old_switch = $$dbh{switch_prepared};
     $$dbh{pg_switch_prepared} = 1;
 
     # Create async statement w/o prepare.
@@ -432,6 +433,8 @@ is ($res, 2, $t);
         $dbh->pg_result();
     };
     is ($@, q{}, $t);
+
+    $$dbh{pg_switch_prepared} = $old_switch;
 }
 
 $dbh->do('DROP TABLE dbd_pg_test5');
--- a/t/08async.t
+++ b/t/08async.t
@@ -26,7 +26,7 @@ if (! $dbh_noerr) {
 $dbh_noerr->{RaiseError} = 0;
 $dbh_noerr->{PrintError} = 0;
 
-plan tests => 116;
+plan tests => 117;
 
 isnt ($dbh, undef, 'Connect to database for async testing');
 
@@ -396,6 +396,43 @@ eval {
 is ($@, q{}, $t);
 $t=q{Database method pg_result returns correct result after execute};
 is ($res, 2, $t);
+
+{
+    my ($sth0, $sth1);
+
+    #
+    # Test that handle_old_async deals correctly with async prepares.
+    # See https://github.com/bucardo/dbdpg/pull/142#issuecomment-4260460925
+    #
+
+    $t=q{Pending async prepare handled correctly by handle_old_async};
+
+    # Start an async prepare.
+    $sth0= $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => PG_ASYNC, pg_prepare_now => 1 });
+
+    # Tell execute that it should prepare on first execution.
+    $$dbh{pg_switch_prepared} = 1;
+
+    # Create async statement w/o prepare.
+    $sth1 = $dbh->prepare('select * from dbd_pg_test5 where t = ?', { pg_async => PG_ASYNC | PG_OLDQUERY_WAIT});
+
+    # Execute. This will call handle_old_async to wait for the result
+    # of the previous async prepare & query and then start a second
+    # async prepare.
+    $sth1->execute(2);
+    eval {
+        # Wait for result.
+        # Without the fix, this will abort with
+        #
+        #  prepared statement "dbdpg_p3993_X" already exists
+        #
+        # as first and second prepare will have used the same
+        # prepare_number because handle_old_async didn't increment
+        # it.
+        $dbh->pg_result();
+    };
+    is ($@, q{}, $t);
+}
 
 $dbh->do('DROP TABLE dbd_pg_test5');
 
--- a/t/08async.t
+++ b/t/08async.t
@@ -408,11 +408,11 @@ is ($res, 2, $t);
     $t=q{Pending async prepare handled correctly by handle_old_async};
 
     # Start an async prepare.
-    $sth0= $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => PG_ASYNC, pg_prepare_now => 1 });
+    $sth0 = $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => PG_ASYNC, pg_prepare_now => 1 });
 
     # Tell execute that it should prepare on first execution.
     $old_switch = $$dbh{switch_prepared};
-    $$dbh{pg_switch_prepared} = 1;
+    $dbh->{pg_switch_prepared} = 1;
 
     # Create async statement w/o prepare.
     $sth1 = $dbh->prepare('select * from dbd_pg_test5 where t = ?', { pg_async => PG_ASYNC | PG_OLDQUERY_WAIT});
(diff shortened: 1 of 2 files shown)


t/08async.t
```diff
@@ -397,6 +397,46 @@ is ($@, q{}, $t);
...
+
+    #
+    # Test that handle_old_async deals correctly with async prepares.
+    # See https://github.com/bucardo/dbdpg/pull/142#issuecomment-4260460925
+    #
+
+    $t=q{Pending async prepare handled correctly by handle_old_async};
+
+    # Start an async prepare.
+    $sth0= $dbh->prepare('select * from dbd_pg_test5 where id = ?', { pg_async => PG_ASYNC, pg_prepare_now => 1 });
```
esabol: Nitpick: Add a space after `$sth0`.

t/08async.t
```diff
@@ -397,6 +397,46 @@ is ($@, q{}, $t);
...
+
+    # Tell execute that it should prepare on first execution.
+    $old_switch = $$dbh{switch_prepared};
```
esabol, line 414: It's a matter of style, and I'm not sure this project has a stance on this particular style, but I personally prefer `$dbh->{switch_prepared}` to `$$dbh{switch_prepared}`.

Ditto on the very next line.

esabol (commented):
I still want to test these changes with some other test scripts (I do that this weekend), but here are some unimportant style comments for now.

t/08async.t
rweikusat: I noticed this but didn't think it was worth another commit.

t/08async.t
rweikusat: Hard references were added to Perl 5 in a way supposed to be compatible with the earlier practice of using symbolic references, that is, variables holding the name of another variable. That's why the mimic symbolic reference syntax. The first rule for working with them is

_A variable holding a reference to an object of a proper type can be used instead of an identifier in an expression._ 

In practice, this means that `$a[0]` is the first element of the array `@a` and `$$a[0]` is the first element of the array `$a` refers to.  This needs one character less than the equivalent `$a->[0]` without being any less clear provided people understand the abovementioned rule. I used to use arrow notation for years because I hadn't really understood it and also, because it looked more like C. But Perl isn't C and `$$a[0]` is - so to say - the native syntax.

rweikusat:
A more general comment: This isn't really a terribly important fix because people hitting this bug (or rather, this trio of bugs) could solve their problem more easily by simply not using async at all as they're not using it in a way where it would benefit them anyhow. 

The idea behind the async API is that one can send queries to the server without waiting for the result. Instead, one will usually have a select/ poll/ epoll/ younameit loop waiting for input on a set of file descriptors with `$$dbh{pg_socket}` being one of them. Once the multiplexing call indicates that input from the server is available, one would call `$dbh->pg_db_ready` to consume it and determine if the query finished. If it did, `$dbh->pg_result` will process the result without blocking. Otherwise, the code would return to the central I/O loop in order to possibly do other stuff until the query has completed.

Invoking `->execute` on a statement handle before a previous async query has finished will simply (`PG_OLDQUERY_WAIT`) block until this query has finished, which renders the whole exercise moot. If blocking until a query finished isn't an issue, one could have avoided using an asynchronous query to begin with.

t/08async.t
esabol: If I didn't think it was worth fixing, I wouldn't have mentioned it. I would appreciate it if you'd fix it.

t/08async.t
esabol: I know all that, tyvm. I still prefer arrow notation because it's more readable (imho). As I said, it's a matter of coding style preference. I defer to @turnstep on the matter.

esabol:
> A more general comment: This isn't really a terribly important fix because people hitting this bug (or rather, this trio of bugs) could solve their problem more easily by simply not using async at all as they're not using it in a way where it would benefit them anyhow.

I disagree. If someone has code that has worked for years (even if it may not be the best written code ever) and that code breaks after updating DBD::Pg to the latest version, I consider that a regression, and I feel all regressions are important to fix.

esabol:
I have tested these changes with [90issue_183.t](https://github.com/user-attachments/files/26878246/90issue_183.txt), and they pass that test as well. The solution looks good to me. I've asked the issue reporter to test as well.

pepl:
> I have tested these changes with [90issue_183.t](https://github.com/user-attachments/files/26878246/90issue_183.txt), and they pass that test as well. The solution looks good to me. I've asked the issue reporter to test as well.

Works

t/08async.t
turnstep: I'm with @esabol on this one, readability is key here. The $$ format looks like a possible typo, while -> is much more clear and explicit. It also chains better. It's also what DBI itself uses, and while we are not beholden to their style, it is a strong influence.

rweikusat: It may look like a typo to you but it's well-defined Perl syntax:

_Anywhere you'd put an identifier (or chain of identifiers) as part of a variable or subroutine name, you can replace the identifier with a simple scalar variable containing a reference of the correct type_
(perldoc perlref)

There's also usually no point 'chaining' arrows. I'assuming this means something like 

`$a->[0]->{honk}->[23]`

which is the same as `$a->[0]{honk}[23]` or `$$a[0]{honk}[23]`.

But while there's little to say in favour of having loads of redundant `>` characters in code, that's not really worth a discussion.

rweikusat:
_[useless use of async]_

_I disagree. If someone has code that has worked for years (even if it may not be the best written code ever) and that code breaks after updating DBD::Pg to the latest version, I consider that a regression, and I feel all regressions are important to fix._

There isn't really anything to "disagree" about here. The facts of the situation are such that the issue only occurs when someone's using the async API for synchronous queries. As that's pointless, simply not using it is the easiest way to avoid this.

esabol (approved):
OK, changes have been made, and all tests pass. Everything looks good to me.

@turnstep : Do you want to review the changes?

turnstep:
Looks good to me. Any objections to a squash commit? I'm fine either way.

esabol:
> Looks good to me. Any objections to a squash commit? I'm fine either way.

I like squashing. The dropdown menu next to the "Merge pull request" button has an option to "Squash and merge", so it can be squashed that way or @rweikusat can squash if he has the time and is amenable.

mauke:
> Invoking `->execute` on a statement handle before a previous async query has finished will simply (`PG_OLDQUERY_WAIT`) block until this query has finished, which renders the whole exercise moot. If blocking until a query finished isn't an issue, one could have avoided using an asynchronous query to begin with.

I don't quite follow. If my workload looks as follows:

1. execute SQL statement 1
2. do local stuff, write files, etc
3. execute SQL statement 2
4. do other local stuff

... and step 2 does not depend on the results of step 1, do I not get better overall runtime by using asynchronous queries? I don't want to block in step 1 because while the DBMS is working, I can do other things (step 2). But if step 2 finishes before SQL statement 1 is done (which can happen sometimes), I do want to block in step 3 before issuing SQL statement 2.

By switching to synchronous queries, I would lose concurrency between step 1 and 2.

rweikusat:
This looks very much like a contrived example patterned after the contrived examples in the man page and one which doesn't really fit the situation at all: If you'd want to _block in step 3 before issuing SQL statement 2_ you'd be calling `$dbh->pg_result` at this point and everything would work fine without the fixes in this PR.

It's still not a particularly sensible use of the async API whose purpose is to avoid blocking for an indeterminate (and possibly even infinite time, as TCP-connections can silently go bad during a period of inactivity) on a single file descriptor because a program might need to do something else at any time aka react to available input on another file descriptor (or set of file descriptors). 

If such blocking is not an issue, executing queries synchronously will make more sense except in pathological cases (such as the one described in the man page). The query transmission time will be negligible even if the server is on a distant continent (ssh connections from England to Japan or the US west coast work fine). If the program can block for an indeterminate amount of time at some particular point of time, the query can be issued then and the step 2 _other stuff_ simply done before the query is issued.

esabol:
I have squashed the PR's commits and merged in https://github.com/bucardo/dbdpg/commit/6793dd5fccaccc277b7b4437a1f8c256566b7db6. Thank you, @rweikusat, for your contribution!

Issue #180, opened 2026-03-26
POD suggests it's impossible to verify server-side SSL certificates


akarelas:
In https://metacpan.org/pod/DBD::Pg , it says:

"There are four valid arguments to the sslmode parameter, which controls whether to use SSL to connect to the database:"

So there are two issues:

1) "There are four" implies to many people that there are not five or more
2) The very important option to verify server SSL certificate (`sslmode=verify-full;sslrootcert=system`) is missing from the list of the four valid arguments.

esabol:
I suspect this POD text was written before PostgreSQL 8.4 was released. Going by the PostgreSQL documentation, 8.4 was the first version to add support for `sslmode=verify-full` and `sslmode=verify-ca`.

DBD::Pg claims to support PostgreSQL 8.0 or higher, so the project can't just add the newer sslmode options to the POD without mentioning they only apply to PostgreSQL version 8.4 or higher. 

Anyone want to draft up what it should say?

akarelas:
Maybe also include instructions for where to find the full list of options, to make the POD future-proof

esabol:
@akarelas : Please take a look at PR #[186](https://github.com/bucardo/dbdpg/pull/186/changes) and let me know what you think. Thanks!

akarelas:
Looks very good, thanks!

PR #186, opened 2026-04-24
Issue #180: Update sslmode documentation

esabol:
This PR updates the `sslmode` connection parameter documentation. It should address issue #180.


Pg.pm
```diff
@@ -1767,9 +1767,9 @@ specific environment variables, and then fall back to hard-coded defaults:
 
 * May also use the aliases C<db> or C<database>
 
-If the username and password values passed via C<connect()> are undefined (as opposed
```
turnstep, line 1770: This seems unrelated to the ssl changes?

turnstep (approved):
Looks good other than the one thing.

Pg.pm
esabol: Yeah, the lines were too long, so I re-wrapped them. I didn't change anything but whitespace.

Do you want me to revert?

turnstep: Yes, a git blame will forever show you as the author. Would rather clean whitespace as a separate commit. But I'm also find with lines over 70 characters long. In the year 2026 we can handle 120+ character lines :)

esabol: OK, reverted and squashed.

Issue #187, opened 2026-04-29
PGresult* leak in `pg_st_deallocate_statement()` when using `PQclosePrepared` (libpq >= 17)


rorsarach:
Background

While investigating a memory growth issue in a derivative PostgreSQL driver based on DBD::Pg 3.10.x, we traced it to a lifecycle bug in `dbd_st_destroy()` and `dbd_st_finish()`. After reviewing DBD::Pg 3.20.0 source code, the SQL `DEALLOCATE` path appears to be fixed by the pointer comparison in `dbd_st_destroy()`. However, the `PQclosePrepared()` path introduced in 3.20.0 for libpq >= 17 has the same class of bug.

This report covers the `PQclosePrepared` path in DBD::Pg 3.20.0.

## What Happens

When `pg_st_deallocate_statement()` takes the `PQclosePrepared` branch, it does:

```c
/* dbdimp.c line 4205 */
imp_dbh->last_result = imp_sth->result
    = PQclosePrepared(imp_dbh->conn, imp_sth->prepare_name);
imp_dbh->result_clearable = DBDPG_FALSE;
```

At this point, both `imp_sth->result` and `imp_dbh->last_result` are overwritten with the result of `PQclosePrepared()`. The previous SELECT result (`R_select`) that was stored in those two pointers is lost — `PQclear(R_select)` is never called.

The reason `R_select` was still in those pointers: `dbd_st_finish()` in 3.20.0 does not touch `imp_sth->result` at all. It only handles async state and calls `DBIc_ACTIVE_off()`. So from execute time to destroy time, `imp_sth->result` keeps pointing to the SELECT result.

## Step-by-step

```
1. $sth->execute()

   Execute stores the result:
     imp_sth->result      = R_select
     imp_dbh->last_result = R_select
     result_clearable     = FALSE

2. All rows fetched.

   dbd_st_fetch() clears DBIc_ACTIVE after the last row.
   imp_sth->result still = R_select (finish() does not touch it).

3. $sth destroyed. prepared_by_us = TRUE, so dbd_st_destroy() calls
   pg_st_deallocate_statement().

4. Inside pg_st_deallocate_statement(), libpq >= 17 branch:

     imp_dbh->last_result = imp_sth->result
         = PQclosePrepared(conn, prepare_name);
     result_clearable = FALSE

   R_select has no remaining reference. PQclear was not called.

5. Back in dbd_st_destroy():

[...]

PR #188, opened 2026-04-29
Fix PGresult leak in PQclosePrepared cleanup path (libpq >= 17)

rorsarach:
In `pg_st_deallocate_statement()`, free the existing sth result before calling `PQclosePrepared()`. #187


esabol:
@rorsarach : The CI tests are failing because of Perl Critic things. Please fix. Make sure you setenv RELEASE_TESTING=1 and AUTHOR_TESTING=1 when you run the tests locally.

I suspect @turnstep will want some other changes to your test script as well in order to fit in better with the project's testing framework. EDIT: Actually, this looks pretty close to the style of most of the project's test files, so maybe not. Good job!

esabol:
@rorsarach : It is still failing a test in t/99_lint.t. Please revise. Make sure you setenv RELEASE_TESTING=1 and AUTHOR_TESTING=1 when you run the tests locally.

esabol:
All CI tests pass now. Thank you, @rorsarach. And, for the record, I have confirmed that the test script fails on DBD::Pg 3.20.2.

dbdimp.c
```diff
@@ -4203,6 +4203,22 @@ static int pg_st_deallocate_statement (pTHX_ SV * sth, imp_sth_t * imp_sth)
...
+     */
+    if (imp_sth->result) {
+        if (imp_sth->result != imp_dbh->last_result) {
+            TRACE_PQCLEAR;
+            PQclear(imp_sth->result);
+        } else {
+            TRACE_PQCLEAR;
+            PQclear(imp_dbh->last_result);
+            imp_dbh->last_result = NULL;
+        }
```
turnstep: I'm a little confused by the logic here. If they ARE equal, we only clear imp_dbh, but if they are NOT equal, we only clear imp_sth? Maybe we want to do something like the other places where we clear before set, like line 3810?

t/13leak.t
```diff
@@ -0,0 +1,225 @@
+#!perl
+
+## Regression tests for the PQclosePrepared result cleanup path (libpq >= 17).
```
turnstep, line 3: I'm not sure we need a whole test file just for a single fix in the code. Generally, we try to narrow the case to a single failing test and put that into an existing file. Maybe that can't be done here, but I think we should try.

t/13pqclose_leak.t
```diff
@@ -0,0 +1,225 @@
...
+plan tests => 9;
+
+isnt ($dbh, undef, 'Connect to database for PQclosePrepared leak regression tests');
+
+## Setup
+
+$dbh->{AutoCommit} = 0;
+
+$dbh->do(q{
+    CREATE TEMP TABLE pqclose_leak_test (
```
turnstep: Tables should start with dbd_pg_ even if they are temporary

t/13leak.t
```diff
@@ -0,0 +1,225 @@
...
+        1;
+    };
+    diag $@ if !$ok && $@;
+
+    my $rss_after = _rss_kb();
+    my $growth_kb = $rss_after - $rss_before;
+    my $limit_kb  = 8 * 1024;
+
+    ok ($ok && $growth_kb < $limit_kb,
+        "RSS growth over 1000 cycles (${growth_kb} kB) is below ${limit_kb} kB limit"
```
turnstep, line 218: Again, this bit is quite clever and a good test, but I'm not sure it's something that needs to be run in CI forever forward once we have determined the leak is fixed. Opinions by others?

dbdimp.c
```diff
@@ -4203,6 +4203,22 @@ static int pg_st_deallocate_statement (pTHX_ SV * sth, imp_sth_t * imp_sth)
...
+        imp_sth->result = NULL;
+    }
+
     imp_dbh->last_result = imp_sth->result
         = PQclosePrepared(imp_dbh->conn, imp_sth->prepare_name);
```
turnstep, line 4217: Checking other places where we might need this, looks like line 5559 could be standardized as the others are.

turnstep (commented):
Thanks for catching this and writing some great tests! We'll get this sorted out and fixed soon.

dbdimp.c
rorsarach: Thanks for flagging that this is confusing. The intent was to avoid calling PQclear twice on the same PGresult: imp_sth->result and imp_dbh->last_result often point to the same object returned from a single PQexecPrepared (the execute path assigns them together as imp_dbh->last_result = imp_sth->result = …). When they are equal, we should call PQclear only once, null out both pointers, and only then hand off to PQclosePrepared to set the new result.

When they are not equal, the current code only frees imp_sth->result. From an ownership perspective, the subsequent imp_dbh->last_result = imp_sth->result = PQclosePrepared(...) overwrites whatever last_result was pointing at on the connection. If last_result still referred to a different, not-yet-freed PGresult, there is in principle the same kind of risk as “clear the sth side but not the last_result side.”

t/13leak.t
esabol: To which test file do you think these tests should be added, @turnstep?

t/13leak.t
esabol: I'm inclined to retain these tests to prevent possible future regressions. The tests are not particularly slow on my system, and they are faster as a whole than some of the existing test files. 

It also provides a framework for future memory leak testing, which could be useful.

I would be in favor of integrating these tests into one of the existing test files, however.

Changes
```diff
@@ -6,6 +6,9 @@ Github items can be found by using: https://github.com/bucardo/dbdpg/issues/XXX
 
 Version 3.20.1  (released April 30, 2026)
 
+ - Fix a memory leak in the libpq 17 PQclosePrepared cleanup path when
```
esabol, line 12: Versions 3.20.1 and 3.20.2 have already been released. So your changes to Changes will need to be revised, @rorsarach. I suggest putting your entry under "Version TBD (released TBD)".

dbdimp.c
esabol: I'm not sure I understand what you're suggesting here, @turnstep.

Line 5559? In which version? Not @rorsarach's, I think.

https://github.com/rorsarach/dbdpg/blob/3cb39cf3b676f1d48dea7908b316bf2eecc7ec97/dbdimp.c#L5559

Do you mean the current master, lines 5555-5589?

https://github.com/bucardo/dbdpg/blob/ac04cc6de29932ea5262519e2f846f477231675c/dbdimp.c#L5555-L5589

turnstep: Yes, I meant the current master. Notice how it has a check for if the sth result is the same as the dbh last_result. We don't need to do that. Just clear and null them both, which simplifies the code and is more bulletproof.

t/13leak.t
turnstep: Well, based on some of the other conversations, perhaps we make this more of a generic memory/leak check test, which just so happens to only have a single item to start with.

esabol:
@rorsarach : The CI tests are suddenly failing. Not sure why. Try rebasing. You'll need to revise your changes to the Changes file again, sorry.

t/13leak.t
esabol: OK, so rename the `t/13pqclose_leak.t` file to something like `t/13memoryleak.t`?

turnstep: Right. Or maybe just 13leak.t to keep with the single word paradigm.

esabol: I like `t/13leak.t`. Much more concise. :smile:

Would you be willing to make that change, @rorsarach ?

Changes
```diff
@@ -4,6 +4,11 @@ RT items can be found by using: https://rt.cpan.org/Public/Bug/Display.html?id=X
 Github items can be found by using: https://github.com/bucardo/dbdpg/issues/XXX
 
 
+Version TBD  (released TBD)
```
esabol: Please remove this line and move your changes entry under `Version 3.20.3  (unreleased)`. Thanks!

t/13leak.t
rorsarach: Should I change t/13pqclose_leak.t to 13leak.t?
I'm happy to make the change.

Changes
rorsarach: My mistake, I thought it was to be changed in the TBD release notes.

Changes
```diff
@@ -9,6 +9,11 @@ Version 3.20.3  (unreleased)
  - Support whitespace in directory names for the test suite.
    [Greg Sabino Mullane]
 
+ - Fix a memory leak in the libpq 17 PQclosePrepared cleanup path when
+     destroying server-side prepared statement handles.
```
esabol: CI tests are still failing. I think it's because the indentation here is wrong? Lines 13 through 15 should line up with the "F" on line 12.

esabol: The CI tests are still failing. :(

In Changes, try joining lines 12 and 13 together into one long line. The test seems to want only one line for the description of each change.

In MANIFEST, you forgot to change `t/13pqclose_leak.t` to `t/13leak.t`.

esabol: @rorsarach : Great! All CI tests pass now. We just need you to reply to @turnstep's unresolved comments above.

rorsarach:
Addressed @turnstep's unresolved review comments:

**1. dbdimp.c cleanup logic (inline comment at the if/else block):** Good catch — the original if/else had a subtle gap: when `imp_sth->result != imp_dbh->last_result`, the old `imp_dbh->last_result` was never freed before being overwritten (potential leak). Simplified it to the clear-before-set pattern like line 3810:
- Clear `imp_sth->result` only if it differs from `imp_dbh->last_result` (avoids double-free when they're the same pointer)
- Unconditionally clear and null `imp_dbh->last_result`
- Null `imp_sth->result`
This handles all pointer combinations cleanly.

**2. Table name (comment on t/13pqclose_leak.t line 32):** The table has always been named `dbd_pg_pqclose_leak_test`, which already starts with `dbd_pg_`. No change needed.

**3. Single test file (comment on t/13leak.t line 3):** Addressed by renaming to `/13leak.t` as a generic memory/leak framework, per your own suggestion ('Or maybe just 13leak.t to keep with the single word paradigm').

**4. RSS growth test in CI forever (comment on t/13leak.t line 218):** I agree with @esabol — the test is fast, prevents future regressions, and provides a reusable framework for future leak tests. I'd like to keep it.

**5. line 5559 standardization (comment on dbdimp.c line 4221):** This points at the `imp_sth->result != imp_dbh->last_result` guard in the existing async result-store path (lines 5576-5579 of current master). That is pre-existing code outside the scope of this PR. Happy to submit a separate follow-up PR to address it if helpful.

[Edited by @esabol to fix formatting.]

rorsarach (commented):
Simplify PQclosePrepared result cleanup in pg_st_deallocate_statement

dbdimp.c
esabol: @turnstep : Is there a specific change you want @rorsarach to make here?

Tbh, it sounds like it might be beyond the scope of this PR. Perhaps we could merge this PR and you can make improvements along the lines of what you have suggested in follow-up commits?

turnstep: Yes, that's fine - I'll fix that up once this PR is complete. Let's not worry about it here.

dbdimp.c
esabol: @turnstep : Have you had the change to look over @rorsarach 's most recent commit? c52ccac198a9f3ea11df6aab507967309b9b1882 It simplifies the logic somewhat.

OK to merge?

turnstep: We always want to clear sth, we all agree on that. But there are times when we do NOT want to clear dbh. That's why there needs to be a `result_clearable` check. If `result_clearable` is false, that means that something else is also looking at that result (e.g. another sth), so we cannot free it (unless that someone else is us!) But if it's not ours, it is safe to move the `imp_dbh->last_result` pointer without freeing it, as it's Somebody Else's Problem. So the code becomes this:
```
if (imp_sth->result) {
  TRACE_PQCLEAR;
  PQclear(imp_sth->result);
}
/* If last_result was from us, ignore result_clearable and set last_result to null */
if (imp_dbh->last_result == imp_sth->result) {
  imp_dbh->last_result = NULL;
}
/* Otherwise, free it unless somebody else has asked us not to */
else if (imp_dbh->last_result && imp_dbh->result_clearable) {
  TRACE_PQCLEAR;
  PQclear(imp_dbh->last_result);
  imp_dbh->last_result = NULL;
}
imp_sth->result = NULL;

```
Took me a little bit to wrap my head around, but I think the above is sound. If something else comes in and sets a new last_result after we do, then we no longer match. We may or may not clear it, depending on if that other process ceded control, but either way, we are going to replace the last_result pointer. Technically, those NULL settings are not needed, but it's good practice.

turnstep: Obviously we will have to clean up all the places we do this to make sure it all works the same way (not in this PR)

esabol: @rorsarach : Would you be willing to change your changes to dbdimp.c to look like the code in @turnstep's post above here?

rorsarach: Sorry, I didn't understand how the code should be modified, based on the chat I thought this piece of code didn't need me to modify it.
Is this code based on this modification?

```
if (imp_sth->result) {
  TRACE_PQCLEAR;
  PQclear(imp_sth->result);
}

/* If last_result was from us, ignore result_clearable and set last_result to null */
if (imp_dbh->last_result == imp_sth->result) {
  imp_dbh->last_result = NULL;
}

/* Otherwise, free it unless somebody else has asked us not to */
else if (imp_dbh->last_result && imp_dbh->result_clearable) {
  TRACE_PQCLEAR;
  PQclear(imp_dbh->last_result);
  imp_dbh->last_result = NULL;
}

imp_sth->result = NULL;
```

rorsarach: Already done in commit c52ccac. The if/else block has been replaced with two independent checks — first clearing imp_sth->result only when it differs from imp_dbh->last_result (to avoid double-free), then unconditionally clearing and nulling imp_dbh->last_result, and finally nulling imp_sth->result. This matches the clear-before-set pattern @turnstep suggested.

turnstep: You still need to be checking `imp_dbh->result_clearable`

rorsarach: Done in commit https://github.com/bucardo/dbdpg/pull/188/changes/0d9016cf2821616efb0f77e4aecc12b387e99c04. The cleanup logic has been revised to match the two-step pattern used in dbd_st_execute (around line 3811):

Free last_result only if it is independently owned (result_clearable is true).
Free sth->result unconditionally.
Why the original code was wrong:

The original if/else had two problems.

First, the guard imp_sth->result != imp_dbh->last_result prevented freeing in the common case. After PQexecPrepared, both pointers share the same PGresult and result_clearable is FALSE. With the guard, neither branch fired, so the PGresult leaked on every prepare/execute/destroy cycle. This is confirmed by t/13leak.t (RSS grows ~37 MB over 1000 cycles on a patched build without the fix).

Second, last_result was freed unconditionally without checking result_clearable. If a second statement (sth_b) had executed after sth_a, last_result would point to sth_b's PGresult with result_clearable = FALSE. Destroying sth_a would then free sth_b's result early, leaving sth_b with a dangling pointer.

How the new code handles each case:

Common case (last_result == sth->result, result_clearable = FALSE): step 1 is skipped; step 2 frees the shared PGresult exactly once.
last_result independently owned (result_clearable = TRUE, different pointer): step 1 frees last_result; step 2 frees sth->result; one free each, no double-free.
Another sth's result in last_result (result_clearable = FALSE, different pointer): step 1 is skipped (protecting the other sth's result); step 2 frees only this sth's result.

esabol: The code in the latest commit still doesn't match exactly what @turnstep suggested, @rorsarach. That seems to be the hold up here on merging this PR. Is your version the same or better?

rorsarach: The code suggested by @turnstep is as follows, I understand as expected is something like this
```
if (imp_sth->result) {
  TRACE_PQCLEAR;
  PQclear(imp_sth->result);
}
/* If last_result was from us, ignore result_clearable and set last_result to null */
if (imp_dbh->last_result == imp_sth->result) {
  imp_dbh->last_result = NULL;
}
/* Otherwise, free it unless somebody else has asked us not to */
else if (imp_dbh->last_result && imp_dbh->result_clearable) {
  TRACE_PQCLEAR;
  PQclear(imp_dbh->last_result);
  imp_dbh->last_result = NULL;
}
imp_sth->result = NULL;
```

When actually tested, this code still has a memory leak. The modified version uses imp_dbh->result_clearable to prevent this problem, like this:
```
/* Free any existing results before overwriting with PQclosePrepared's return value */
if (imp_dbh->last_result && imp_dbh->result_clearable) {
    TRACE_PQCLEAR;
    PQclear(imp_dbh->last_result);
    imp_dbh->last_result = NULL;
}
if (imp_sth->result) {
    TRACE_PQCLEAR;
    PQclear(imp_sth->result);
    imp_sth->result = NULL;
}
```
Or does @turnstep have any other suggestions?

turnstep: For this solution:

```
/* Free any existing results before overwriting with PQclosePrepared's return value */
if (imp_dbh->last_result && imp_dbh->result_clearable) {
    TRACE_PQCLEAR;
    PQclear(imp_dbh->last_result);
    imp_dbh->last_result = NULL;
}
if (imp_sth->result) {
    TRACE_PQCLEAR;
    PQclear(imp_sth->result);
    imp_sth->result = NULL;
}
```

...the problem is that we might call PQclear twice on the same result, which is not good. This can happen when imp_dbh->last_result == imp_sth->result. Yes, we can still hope that result_clearable is always set to false in that case, but it is not safe to rely on that completely.

Hmmm...walking through the code, maybe it is safe, as we do seem to have 100% coverage of always making sure that result_clearable is set to false when we assign two pointers to the same PGresult.

So now that I've talked out loud through all of that, I'm okay with the solution above. We just have to make sure we are fastidious about setting result_clearable, and have a standard way to clear out old results. That standard way pops up enough to *almost* warrant a separate function or macro, but there is still enough variations to not make it worth it, for now. I will mull on that.

esabol: OK, sounds good. @turnstep, shall we merge this while you're mulling over more widespread improvements?

Issue #192, opened 2026-05-02
clang build warning


esabol:
Debian clang version 19.1.7
CC="clang -Werror"

```
clang -Werror -c  -I/usr/include/postgresql -I/usr/local/lib/perl5/site_perl/5.42.2/x86_64-linux-gnu/auto/DBI -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2 -DPGLIBVERSION=170009 -DPGDEFPORT=5432 -DHAS64BITLO -O2  -DPERL_EXTMALLOC_DEF -Dmalloc=Perl_malloc -Dfree=Perl_mfree -Drealloc=Perl_realloc -Dcalloc=Perl_calloc -DVERSION=\"3.20.2\" -DXS_VERSION=\"3.20.2\" -fPIC "-I/usr/local/lib/perl5/5.42.2/x86_64-linux-gnu/CORE"   dbdimp.c
dbdimp.c:3996:46: error: passing 'unsigned char *' to parameter of type 'char *' converts between pointers to integer types where one is of the unique plain 'char' type and the other is not [-Werror,-Wpointer-sign]
 3996 |                     type_info->dequote(aTHX_ value, &value_len); /* dequote in place */
      |                                              ^~~~~
1 error generated.
make: *** [Makefile:362: dbdimp.o] Error 1
```

Does this naïve fix seem reasonable to you, @turnstep ?

```
--- /path/to/my/git/dbdpg/dbdimp.c   2026-05-02 16:37:01.574283000 -0400
+++ ./dbdimp.c  2026-05-02 17:23:11.722172000 -0400
@@ -3994,5 +3994,5 @@
                 if (type_info) {
                     STRLEN value_len;
-                    type_info->dequote(aTHX_ value, &value_len); /* dequote in place */
+                    type_info->dequote(aTHX_ (char *)value, &value_len); /* dequote in place */
                     /* For certain types, we can cast to non-string Perlish values */
                     switch (type_info->type_id) {
```

esabol:
Dockerfile used for testing:
```
ARG version=5.42
FROM perl:${version}

ARG version
ARG DBDPG_REPO=https://github.com/bucardo/dbdpg

LABEL description="DBD::Pg Testing (Perl ${version})"
[...]

PR #193, opened 2026-05-02
Fix a clang build warning, add clang builds to CI, and add -Werror to all CI builds

esabol:
Report: https://github.com/bucardo/dbdpg/issues/192
Reported-by: Ed Sabol, Peter Dyballa
Co-authored-by: Ed Sabol

This fixes a build warning reported by the clang compiler, adds a couple builds using clang to the CI workflow, and adds `-Werror` to all CI workflow builds.

This will cause the CI clang jobs to fail initially, but then I'll commit a separate change to dbdimp.c which will fix them.

EDIT: The CI jobs all passed after the second commit with the change to dbdimp.c.

### DIFF
--- a/.github/workflows/with_pg.yaml
+++ b/.github/workflows/with_pg.yaml
@@ -18,6 +18,15 @@ jobs:
       matrix:
         perl: [ '5.30', '5.36', '5.40', 'latest' ]
         postgres: [ '11', '13', '15', '17', 'latest' ]
+        compiler: [ 'gcc' ]
+        include:
+          # Add extra jobs for building with clang
+          - perl: 'latest'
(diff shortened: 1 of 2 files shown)


esabol:
@turnstep: [As requested](https://github.com/bucardo/dbdpg/issues/192#issuecomment-4366401590), I have changed `value` in `dbdimp.c` from an `unsigned char *` to a `char *`. Although the changes are considerably longer, this change does simplify the code. All the CI tests still pass. Tested with gcc 15.2 in addition to the CI tests. I agree that this is a better solution. Ready for review/merge.

turnstep (approved):

PR #195, opened 2026-05-22
Return 0 for $sth->row() even when no tuples returned

turnstep:
Per issue 194


esabol (approved):
Simple change that looks correct to me. Good tests added for coverage. All CI tests pass. Thanks!

PR #196, opened 2026-05-27
PQclear macros

turnstep:
Simplify repeated PQclear code for both database and statement handles with two new macros.

Rename result_clearable to result_shared

### DIFF
--- a/dbdimp.h
+++ b/dbdimp.h
@@ -49,7 +49,7 @@ struct imp_dbh_st {
     bool    client_encoding_utf8; /* is the client_encoding utf8 last we checked? */
 
     PGresult  *last_result;     /* PGresult structure from the last executed query (can be from imp_dbh or imp_sth) */
-    bool      result_clearable; /* Is it alright to call PQclear on last_result? (statements handles set it to false */
+    bool      result_shared;    /* Are more than one things pointing to this PGresult? */
     imp_sth_t *do_tmp_sth;      /* temporary sth to refer inside a do() call */
 };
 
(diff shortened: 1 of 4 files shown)


dbdimp.h
```diff
@@ -49,7 +49,7 @@ struct imp_dbh_st {
     bool    client_encoding_utf8; /* is the client_encoding utf8 last we checked? */
 
     PGresult  *last_result;     /* PGresult structure from the last executed query (can be from imp_dbh or imp_sth) */
-    bool      result_clearable; /* Is it alright to call PQclear on last_result? (statements handles set it to false */
+    bool      result_shared;    /* Are more than one things pointing to this PGresult? */
```
esabol: Please change `Are more than one things` to `Is more than one thing`.

esabol (commented):
Other than a minor nit to pick with the grammar in one of the comments, I love everything about this PR. The macros make the code easier to understand and less error prone in future revisions.

However, I wish you hadn't altered the test in 13leak.t in direct repo commits. It's hard to tell what was changed there because the file has been renamed, also. Ideally, the 13leak.t -> 5leak.t and related changes should have been done after this PR was merged. I'd like to test this branch with the original 13leak.t, tbh, but I don't have time to do that right now.

turnstep:
> However, I wish you hadn't altered the test in 13leak.t in direct repo commits

Fair point. I was cleaning up and refactoring and got carried away. However, I did confirm that the existing 5leak has the same behavior as the original one - if you take out the missing PQclears, it fails as it should, and passes once they are added back in.

esabol (commented):
Looks good!

esabol (approved):

PR #201, opened 2026-07-01
Use new string buffer system, to enable arrays rather than linked list.

turnstep:
Based on the work from https://github.com/bucardo/dbdpg/pull/21

Author: Matt Tyson
Co-authored-by: Greg Sabino Mullane


turnstep:
Looking for reviews, this is a substantial bit of work.

dbdimp.c
```diff
@@ -252,7 +346,7 @@ int dbd_db_login6 (SV * dbh, imp_dbh_t * imp_dbh, char * dbname, char * uid, cha
         connect_string_size += strlen("user='' ") + 2*strlen(uid);
     if (*pwd)
         connect_string_size += strlen("password='' ") + 2*strlen(pwd);
-    New(0, conn_str, connect_string_size+1, char); /* freed below */
+    New(0, conn_str, connect_string_size+1, char); /* freed in dbd_st_destroy */
```
esabol: The change to this comment appears to be wrong. It's freed on line 420, which is "below" and not in `dbd_st_destroy`. Revert?

https://github.com/bucardo/dbdpg/blob/24a6a7a11b33e72f368343613ce442c50e1391ae/dbdimp.c#L420

dbdimp.c
```diff
@@ -2673,10 +2726,17 @@ int dbd_bind_ph (SV * sth, imp_sth_t * imp_sth, SV * ph_name, SV * newvalue, IV
         phnum = atoi(name);
         if (phnum < 1 || phnum > imp_sth->numphs)
             croak("Cannot bind unknown placeholder %d (%s)", phnum, neatsvpv(ph_name,0));
-        for (x=1,currph=imp_sth->ph; NULL != currph; currph = currph->nextph,x++) {
-            if (x==phnum)
+        found = 0;
+        for (int p=0; p < ph_array_count(imp_sth); p++) {
```
esabol: Can't you just check that `phnum > 0 && phnum <= ph_array_count(imp_sth)` and then just do this:

`currph = ph_array_element(imp_sth, phnum - 1);`

and eliminate the `for`, making it O(1) instead of O(n)?

dbdimp.c
```diff
@@ -1305,43 +1396,41 @@ SV * dbd_st_FETCH_attrib (SV * sth, imp_sth_t * imp_sth, SV * keysv)
...
             }
             retsv = newRV_noinc((SV*)arr);
         }
         else if (strEQ("pg_numbound", key)) {
-            ph_t *currph;
-            int i = 0;
-            for (currph=imp_sth->ph; NULL != currph; currph=currph->nextph) {
-                i += NULL == currph->bind_type ? 0 : 1;
+            int p, num = 0;
+            for (p=0; p < ph_array_count(imp_sth); p++) {
```
esabol: Can the `int p` declaration on line 1428 be moved into the `for` loop initializer on line 1429? I don't think `p` is used after the `for` loop. I think this is the only `for` loop where you didn't already do that. :smile:

esabol (commented):
Just a few inline comments...

Since this does a lot of memory allocating, it might be worth doing some valgrind checks?

dbdimp.c
turnstep: Wow, that is so much easier, great spot! My brain was way too used to making that for loop.

dbdimp.c
turnstep: Done! I did start sneaking some more C99 inline declarations in for this PR. I don't see any reason at this point to avoid them.

dbdimp.c
turnstep: Yep - I meant this for the *other* freed below!

turnstep:
Valgrind is a good idea, but I was not able to get it working last I tried. That was many years ago, so if I get some free time I will give it another go for 2026.

esabol (commented):
At the risk of being premature, it looks good to me.

esabol (approved):

esabol (approved):
Looks even better!

PR #208, opened 2026-08-25
dbdpg_test_setup.pl: Make port discovery more robust

esabol:
This PR makes port discovery more robust in `dbdpg_test_setup.pl`. `ss` doesn't exist on macOS and *BSD systems. We need to use `netstat` on such systems. Also, `netstat` on macOS and *BSD systems (but not on Linux, apparently) puts a period (instead of a colon) before the port number. Finally, Google Gemini insists there can be a tab (or possibly nothing) after the port number (instead of a space) in some cases.

### DIFF
--- a/t/dbdpg_test_setup.pl
+++ b/t/dbdpg_test_setup.pl
@@ -497,15 +497,15 @@ version: $version
         $info = '';
         my $evalok = 0;
         eval {
-            $info = $^O =~ /Win32/ ? qx{netstat -anp TCP}: qx{ss -QanO 2>&1};
+            $info = $^O =~ /(?:Win32|darwin|bsd|dragonfly)/ ? qx{netstat -anp TCP 2>&1} : qx{ss -anO 2>&1};
             $evalok = 1;
         };
         if (!$evalok or ! defined $info) {
-            Test::More::diag "ss call to determine open port failed, trying port $testport\n";
+            Test::More::diag "Unable to determine open port, trying port $testport\n";
         }
         else {
             {
-                last if $info !~ /:$testport /m;
+                last if $info !~ /^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.]$testport\s/im;
                 last if ++$testport > $TEST_PORT_MAX;
                 redo;
             }


t/dbdpg_test_setup.pl
```diff
@@ -497,15 +497,15 @@ sub connect_database {
         $info = '';
         my $evalok = 0;
         eval {
-            $info = $^O =~ /Win32/ ? qx{netstat -anp TCP}: qx{ss -QanO 2>&1};
+            $info = $^O =~ /(?:Win32|darwin|bsd|dragonfly)/ ? qx{netstat -anp TCP} : qx{ss -QanO 2>&1};
             $evalok = 1;
         };
         if (!$evalok or ! defined $info) {
             Test::More::diag "ss call to determine open port failed, trying port $testport\n";
```
turnstep: Maybe just "Unable to determine open port, trying port $testport"

t/dbdpg_test_setup.pl
```diff
@@ -497,15 +497,15 @@ sub connect_database {
...
         }
         else {
             {
-                last if $info !~ /:$testport /m;
+                last if $info !~ /[:.]$testport(?:\s|$)/m;
```
turnstep: Definitely do not want $, as we don't want peer ports. Doesn't solve netstat though, will ponder

t/dbdpg_test_setup.pl
```diff
@@ -497,15 +497,15 @@ sub connect_database {
         $info = '';
         my $evalok = 0;
         eval {
-            $info = $^O =~ /Win32/ ? qx{netstat -anp TCP}: qx{ss -QanO 2>&1};
+            $info = $^O =~ /(?:Win32|darwin|bsd|dragonfly)/ ? qx{netstat -anp TCP} : qx{ss -QanO 2>&1};
```
turnstep: Ideally we filter out ipv6 as well, as it can lead to false positives, but frustratingly all the netstats I'm looking at (win,osx,nix) have slightly different flags.

t/dbdpg_test_setup.pl
turnstep: After some experimentation, I came up with this:

```
/\b\d+\.\d+\.\d+\.\d+[:.](\d+)\s\D/
```
This catches all the various outputs

t/dbdpg_test_setup.pl
esabol: Done in latest version.

t/dbdpg_test_setup.pl
esabol: `\D` will match `\s` too. Is that what you want?

turnstep: Hmm, that's true. Since both give multi-column output, we could do:

```
/\b\d+\.\d+\.\d+\.\d+[:.](\d+)\s+[0-9\*]/
```

Also, ss should be:
```
ss -Qanof inet
```

esabol: I just discovered that the existing `ss -QanO` doesn't work on my RHEL 8.10 system!

I get `ss: invalid option -- 'Q'`, so `ss -Qanof inet` would also be a no-go. `ss -anof inet` works. Can we live without the `Q`?

Re: the regex, are you trying to only match the first (local) address column?

My Mac's `netstat -anp` output has lines like this:
```
tcp6       0      0  *.17500                *.*                    LISTEN
tcp4       0      0  *.17500                *.*                    LISTEN     
```
If `$testport` were 17500, we'd want to not use that port number, right?

Maybe `/\b(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)\s+[0-9\*]/`?

turnstep: We really need to filter out the udp, as it can give false positives.

Lack of "Q" is fine.

On further thought, maybe the udp has to stay. Just looked at Win32, which does not support  `-f inet`  - and the output is different (sigh). I think all of them at this point look close enough that maybe we can try a anchored regex like so?:

```
/^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)/i
```
(Leading whitespace and case insensitive needed for Win32. Non greedy regex means we should always grab the first  address we find, so I think that accounts for everything)

esabol: I've removed the `Q` from the `ss` command.

Is there a reason why the `ss` command has `2>&1`, but the `netstat` command does not? Is that because of Windows?

If you add a `^` to the regex, then I think you need to add the `m` modifier at the end, so that should be `/^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)/im`.

However, I'm pretty sure we need a `\s` at the end of the regex, no? Otherwise, we'll match port 12345 when we're testing for port 1234. So that becomes:

`/^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)\s/im`

On my RHEL 8.10 system:
```
> perl -E 'my $info = qx{ss -anO 2>&1}; if ($info =~ /^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)\s/im) { say "port $1" } else { say "No match" }'
port 44017
```
On my Mac:
```
> perl -E 'my $info = qx{netstat -anp TCP 2>&1}; if ($info =~ /^\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)\s/im) { say "port $1" } else { say "No match" }'
port 60672
```
I don't have access to a Windows system to test on.

I've committed these changes to this PR. Please review.

turnstep: Windows supports 2>&1 so we can leave it in place everywhere.

That regex seems to work for Win32. Here's some sample output from Windows netstat -anp tcp for reference: 

```
Active Connections

  Proto  Local Address          Foreign Address        State
  TCP    0.0.0.0:123            0.0.0.0:0              LISTENING
  TCP    192.168.1.200:49428    101.283.5.133:443      ESTABLISHED
  TCP    192.168.1.200:50388    112.216.122.127:443    ESTABLISHED
  TCP    192.168.128.1:139      0.0.0.0:0              LISTENING
```

turnstep (approved):

t/dbdpg_test_setup.pl
esabol: Just to confirm, my test with the Win32 output you provided passes:

```
perl -E 'my $info = "Active Connections\n\n  Proto  Local Address          Foreign Address        State\n  TCP    0.0.0.0:123            0.0.0.0:0              LISTENING\n  TCP    192.168.1.200:49428    101.283.5.133:443      ESTABLISHED\n  TCP    192.168.1.200:50388    112.216.122.127:443    ESTABLISHED\n  TCP    192.168.128.1:139      0.0.0.0:0              LISTENING\n"; if ($info =~ /^(\s*tcp.+?(?:\d+\.\d+\.\d+\.\d+|\*)[:.](\d+)\s)/im) { say "$1: port $2" } else { say "No match" }'
  TCP    0.0.0.0:123 : port 123
```

I've added " 2>&1" to the `netstat` command as well. Once the CI tests pass, I'll merge the latest version. Thanks for your assistance!

PR #209, opened 2026-08-28
Cleanup the data_sources function and its tests.

turnstep:
Allow the arg to be a string or a hashref.

Filter out lone 'Pg' to let DBI and $dbh calls work identically.

Add 'postgres' to the list of perlcritic stopwords (how was that not already there!)

### DIFF
--- a/Pg.pm
+++ b/Pg.pm
@@ -1895,7 +1895,9 @@ Implemented by DBI, no driver-specific impact.
 =head3 B<data_sources>
 
   @data_sources = $dbh->data_sources($attr);
+  @data_sources = $dbh->data_sources(\%attr);
   @data_sources = DBI->data_sources('Pg', $attr);
+  @data_sources = DBI->data_sources('Pg', \%attr);
 
 Returns a list of available databases as DBI connection strings.
 
--- a/Pg.pm
+++ b/Pg.pm
@@ -215,22 +215,43 @@ use 5.008001;
 
     use strict;
 
-    ## Returns an array of formatted database names from the pg_database table
+    ## Returns a list of formatted database names from the pg_database table
+    ## Each is prefixed with 'dbi:Pg:dbname=' making it suitable to pass to connect()
+    ## This function can be called via DBI or $dbh
     sub data_sources {
 
         my $drh = shift;
         my $extra_conninfo = shift || '';
-        $extra_conninfo =~ s/^([^;])/;$1/;
+        my $ref = ref $extra_conninfo;
+
+        if ($ref eq 'HASH') {
+            if (%$extra_conninfo) {
+                $extra_conninfo = ';' . join ';' => map { "$_=$extra_conninfo->{$_}" } sort keys %$extra_conninfo;
+            }
+            else {
+                $extra_conninfo = '';
+            }
+        }
+        elsif ($ref eq '') {
+            $extra_conninfo = '' if $extra_conninfo eq 'Pg';
+            $extra_conninfo =~ s/^([^;])/;$1/;
+        }
+        else {
+            die "data_sources(): unsupported argument type $ref: should be string or a hashref\n";
+        }
 
         my $connstring = 'dbname=postgres';
         if ($ENV{DBI_DSN}) {
-            ($connstring = $ENV{DBI_DSN}) =~ s/dbi:Pg://i;
+            ($connstring = $ENV{DBI_DSN}) =~ s/\s*dbi:Pg://ig;
         }
+        $connstring =~ s/^;//;
 
-        my $dbh = DBD::Pg::dr::connect($drh, $connstring) or die 'Could not connect to the database';
+        my $dbh = DBD::Pg::dr::connect($drh, $connstring)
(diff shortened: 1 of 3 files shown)


Pg.pm
```diff
@@ -1873,18 +1894,20 @@ Implemented by DBI, no driver-specific impact.
 
 =head3 B<data_sources>
 
-  @data_sources = $dbh->data_sources('Pg', \%attr);
-  @data_sources = DBI->data_sources('Pg', \%attr);
+  @data_sources = $dbh->data_sources($attr);
+  @data_sources = DBI->data_sources('Pg', $attr);
```
esabol, line 1899: This doesn't seem to match the syntax of `$dbh->data_sources(...)` or `DBI->data_sources(...)` that I see in the DBI documentation? I see no indication there that the `attr` argument can be a scalar string?

References:
https://metacpan.org/pod/DBI#data_sources
https://metacpan.org/pod/DBI#data_sources1

This isn't a method I've ever used, tbh. Sorry if I'm missing some historical context.

The DBI documentation seems to prefer `\%attr`. I do as well.
Let's say the `data_sources` method _can_ take either a hashref or a scalar string... For the documentation, I would write that like this:
```perl
  @data_sources = $dbh->data_sources($attr);
  @data_sources = $dbh->data_sources(\%attr);
  @data_sources = DBI->data_sources('Pg', $attr);
  @data_sources = DBI->data_sources('Pg', \%attr);
```

t/04misc.t
```diff
@@ -457,63 +457,107 @@ dbdpg: Begin _sqlstate
...
-SKIP: {
+$t=q{DBI method data_sources() works when no args and DBI_DRIVER is 'Pg'};
+{ local $ENV{DBI_DRIVER} = 'Pg'; @sources = DBI->data_sources(); }
+like ((join ' ' => @sources), $expected, $t);
+
+$t=q{DBI method data_sources() works when first arg is empty string and DBI_DRIVER is 'Pg'};
+{ local $ENV{DBI_DRIVER} = 'Pg'; @sources = DBI->data_sources(''); }
+like ((join ' ' => @sources), $expected, $t);
 
-    $t='DBI method data_sources() handles database names with spaces';
```
esabol, line 493: You've removed this test? Without it, I think there is no longer a test that exercises a non-trivial database name that requires quotes.

Pg.pm
esabol: What the DBI documentation says DBD::Pg should support is this:
```perl
  @data_sources = $dbh->data_sources();
  @data_sources = $dbh->data_sources(\%attr);
  @data_sources = DBI->data_sources('Pg');
  @data_sources = DBI->data_sources('Pg', \%attr);
```

turnstep: I'm fine with tweaking the docs like that. The whole function is poorly documented and frankly, poorly implemented (if implemented at all) inside other DBDs. At some point we blazed our own path - others use the args as hints to the actual connection, while we use it to modify the returned list. Way too late to change it now, of course, but I think our implementation makes more sense, as one can just use DBI_DSN if you need to do the initial connection with different args.

t/04misc.t
turnstep: Yeah, I decided to leave this out as all we are really doing with that is testing quote_ident(), which is not our purview. It's also a fairly heavy thing for a test to do (create and destroy a whole database) for what is a seldom-used function.

turnstep:
The implicit filtering of the $dbh->data_sources('Pg') has some interesting history - apparently, our tests should have been failing those tests, as it used to append that Pg to the end of the returned dsns, but our regex had a \b at the end, so it was not detected! I made a carve out for a bare 'Pg' to work, and fixed up the regex as well. Could still break if the database template0 does not exist, but nobody should ever be doing that. :)

Pg.pm
turnstep: Looks like we used to append to the connection string too back in 2007, then added to the returned DSNs in 2008, then removed from the connstring this year. I'm open to reverting that behaviour, but seems a little pointless as that's what the DBI_DSN env is for.

t/04misc.t
esabol: OK, those are good reasons to jettison that test then.

esabol:
> The implicit filtering of the $dbh->data_sources('Pg') has some interesting history - apparently, our tests should have been failing those tests, as it used to append that Pg to the end of the returned dsns, but our regex had a \b at the end, so it was not detected! I made a carve out for a bare 'Pg' to work, and fixed up the regex as well. Could still break if the database template0 does not exist, but nobody should ever be doing that. :)

Well, `$dbh->data_sources('Pg')` isn't valid DBI syntax. The driver string argument is only valid when used as a class method: `DBI->data_sources('Pg')`. I can see it makes the implementation easier to handle both class and $dbh method calls with the same code, however.

Pg.pm
esabol: Ah, I figured there must be some kind of historical reason to allow a scalar `$attr` argument.

